PR #84, opened 2014-09-11
Switch account

OriPekelman:
My excuses for this huge, unclean pull request

Added Features:

 * locally build symfony projects
 * allow switching between multiple login accounts (.platform becomes a symlink to a per account setting file in the home directory of the form .platform_https___marketplace.commerceguys.com_ori_pekelman.com)
 * Allow setting market place url through  exporting $PLATFORM_ACCOUNTS_SITE environnement variable (useful for running in dev mode)
 * Allow setting the ssh key for git through the exporting $PLATFORM_IDENTITY_FILE environnement variable (necessary for multiple accounts...) and $GIT_SSH={CLI_ROOT}/platform-git
 * Tag commands as either local or remote and color them nicely

Modifications:
* make some commands unprivileged, currently in a vairable in Application.php ($unprivilegedCommands) probably needs to be handled by each command
* rename and expose the login command, allow passing username in the commandline 
* make ssh keys workflow more userfriendly, detect more cases (no keys on server, no keys locally). When no keys are there list local keys.
* rename sshkeys commands to make them coherenet (this should happen for all commands, no mix of plural and singular.) 

CAVEAT
Extremly lightly tested by hand. Only tested locally on OS X. No automatic testing at all.


pjcdawkins:
Closing this - I have added some better SSH key UX, and #241 allows account switching via a PLATFORM_CLI_SESSION_ID environment variable (default 'default')

PR #86, opened 2014-09-11
Local tunnel

OriPekelman:
Hacky and unfinished.

BAsically what it does is open tunnel(s) for the relations (currently hardcoded).
I put local.platform.sh 127.0.0.1 in my /etc/hosts file but it would be much cooler to resolve local.platform.sh to 127.0.0.1 through dns...
There is some stuff to be done around configuration still for Drupal and probably Symfony (solr config in db?).

```text
A host has been configured as local.platform.sh

A tunnel to environment mybranch has been created (ssh -N -L 18080:solr.internal:8080 -L 16379:redis.internal:6379 -L 13306:database.internal:3306 pxbzj2d3rq6rc-mybranch@ssh.eu.platform.sh)
Connect to solr through local.platform.sh:18080Connect to redis through local.platform.sh:16379Connect to database through local.platform.sh:13306
Set your settings to connect local.platform.sh: for mysql (do not use localhost !)

You can see the site on http://local.platform.sh:8000
```

### DIFF
--- a/bin/mysql-sync
+++ b/bin/mysql-sync
@@ -1,0 +1,415 @@
+#!/usr/bin/php
+
+<?php
+
+/**
+ * Usage:
+ * Example:
+ * mysql-sync root:password@localhost:3308/voeb root:password@localhost:3306/v json|verbose
+ */
+
+mb_internal_encoding("UTF-8");
+
+// print_r($argv);
+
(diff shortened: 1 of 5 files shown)


OriPekelman:
This is stale most everything was integrated somewhere else. This is now basically a documentation issue.  We probably also need better output for `platform relationships` will open a ticket for that.

Issue #170, opened 2014-11-20
platform-cli seems to cache project list


OriPekelman:
I created a new project when typing just `platform` I did not see the new project in the list.
When I type `platform projects` the new project appeared
When typing now `platform` I get the correct list

pjcdawkins:
In 1.x and 2.x, the projects list expires after 1 hour, and the environments list after 10 minutes. And it's (still) invalidated if you run `platform projects` or `platform environments`, respectively.

2.x has the potential for proper HTTP caching (it would be a simple config switch), if it is implemented on the API side.

OriPekelman:
Good enough. And as some remember I will be extatic if we set on the API side the etag to the commit hash.

PR #172, opened 2014-11-20
Add HTTP caching.

pjcdawkins:
Should solve #170 

- [x] Deploy new cache changes (remove Expires, add ETag revalidation) on the API side for the projects list
- [ ] Support ETag-based HTTP caching on the API side for environments
- [ ] Write upgrade hook for moved session info

Questions:
 * where do we best store session data and the cache? I chose `~/.platformsh/session.yml` and `~/.platformsh/cache`, both removed on logout

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,13 +5,14 @@
     "require": {
         "cocur/slugify": "~1.0",
         "commerceguys/guzzle-oauth2-plugin": "~1.0",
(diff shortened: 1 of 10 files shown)


pjcdawkins:
Closing as cache options are built into the client library which will be used in v2 (#241)

Issue #179, opened 2014-11-26
Implement a 'tree id' to save build time


pjcdawkins:
Fingerprint the files inside the application (somehow), and skip building if a matching build already exists.

PR #193, opened 2014-12-04
Archive and re-use old builds if the app's files have not changed

pjcdawkins:
Addresses my #179.

 - [x] is there a way we can make the tree id match Platform's? [not really, no]
 - the tar/untar commands need cross-platform testing:
   - [x] OS X (Yosemite)
   - [x] Ubuntu (Trusty)
   - [x] Fedora (20)
   - [x] Other Unix (FreeBSD 10)
   - [ ] Windows 7
   - [ ] Windows 8
 - [x] there is no point keeping old build directories with this (not sure there was any point in the first place)
 - [x] refactor and modify the 'clean' command so that it deletes the 'slugs'
 - [ ] write automated tests

### DIFF
--- a/src/Command/ProjectBuildCommand.php
+++ b/src/Command/ProjectBuildCommand.php
@@ -21,7 +21,17 @@ class ProjectBuildCommand extends PlatformCommand
                 'abslinks',
                 'a',
                 InputOption::VALUE_NONE,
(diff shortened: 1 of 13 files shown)


pjcdawkins:
Merging to help test on Windows, further tests/fixes can be in a follow-up

PR #198, opened 2014-12-11
Ensure a Git remote named 'platform'.

pjcdawkins:
This could allow users to have an 'origin' remote set to another URL.

[WIP]


DuaelFr:
For information: having to deal with more than one remote repository can be made with only one remote identifier as explained here http://stackoverflow.com/a/4255934/766215
I just made the test and it works well. Of course, the two remotes have to be absolutely synchronized.

pjcdawkins:
I think we want two identifiers for the same remote, rather than one identifier for two remotes. I want consistent documentation, and a reliable remote name that can be used by other commands. I want to be able to advise a CLI user: `git push platform <branch>` and to know that it'll work. But users will also be used to working with `origin`.

PR #203, opened 2014-12-19
Ability to modify access on environment.

kotnik:


kotnik:
@pjcdawkins thanks for the review!

PR #204, opened 2014-12-21
Add an installer script.

pjcdawkins:
Script and/or install instructions need testing to support:
 - [x] Ubuntu (Bash)
 - [x] OS X (ZSH / Bash)
 - [x] Fedora (Bash)
 - [x] Other *nix (FreeBSD 10 works, but requirements/dependency checks are probably not useful enough)
 - [x] MinGW (works but has some errors, and no autocomplete)
 - [ ] Cygwin
 - [ ] PowerShell

Test with
```sh
curl -sfS https://raw.githubusercontent.com/pjcdawkins/platformsh-cli/installer/installer.sh | bash
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The **Platform.sh CLI** is the official command-line interface for [Platform.sh]
 
 ### Installation
 
-* On Windows, [install Composer globally](https://getcomposer.org/doc/00-intro.md#globally).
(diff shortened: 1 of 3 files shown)


pjcdawkins:
This is a hack around Composer. Box project integration will be nicer for the end user (even though it's more complicated)

Issue #212, opened 2015-01-19
Split up local shared resources by application (RFC)


pjcdawkins:
Current situation:
 * everything from `shared` gets symlinked into `sites/default` for all Drupal builds

Problems:
 * Symfony applications should be allowed to have a `config_dev.yml` in `shared`, symlinked to `app/config`, but at the moment that would also get symlinked to sites/default for the Drupal builds
 * perhaps there's more than one Drupal application, but the user wants/needs to have different settings.local.php files for each

Possible solution:
 * Detect/create `shared/<appname>`, and if it exists, use it instead of `shared`

PR #213, opened 2015-01-19
Refactoring for local project/filesystem info

pjcdawkins:
Addresses #212


pjcdawkins:
In a multi-application project, the `shared` files will be placed in `<projectRoot>/shared/<appName>`

In single-application projects, the shared files stay in `<projectRoot>/shared`

PR #216, opened 2015-01-26
Show more iformation in the welcome message for current project

andyg5000:
It would be helpful to have the project URI and ID displayed whenever running the platform command within the context of a project.


pjcdawkins:
Yes you're right - thanks for this

PR #223, opened 2015-01-31
Display the command name and aliases in the command's help.

pjcdawkins:
Again, it's not my fault... you do really need this much code just to override a small aspect of Symfony's behaviour.


pjcdawkins:
screen shot 2015-02-01 at 09 48 58

Issue #232, opened 2015-02-06
Support .platform.app.local.yaml, to define CLI-specific build hooks


pjcdawkins:

dmsmidt:
This would be really nice.
Maybe it could even be extended into .platform.app.ENVIRONMENT.yaml? But I guess this is out of the scope of the CLI tool.

pjcdawkins:
I *think* that wouldn't happen, because (apart from some metadata - access control, "variables", etc.) environments should be identical, only affected by changes tracked in git. When you merge env2 into env1, you should expect env1 to be identical.

pjcdawkins:
@dmsmidt ^

and an example .platform.app.local.yaml file might be:
```yaml
hooks:
    build: |
        mv profiles/my_profile/my_file.txt my_file.txt
    deploy: |
        drush -y updb
        drush cc drush
        drush -y fra
```

and the output would look like:
```
Running local build hook(s)
Saving build archive...
Running local deploy hook(s)
No database updates required                                                                                        [success]
'all' cache was cleared.                                                                                            [success]
Finished performing updates.                                                                                        [ok]
'drush' cache was cleared.                                                                                          [success]
Current state already matches defaults, aborting.                                                                   [ok]
```

pjcdawkins:
Closing in favour of #243, which will be harder to code... but easier for the end user

PR #237, opened 2015-02-12
Support build/deploy hooks in a .platform.app.local.yaml file

pjcdawkins:
Addresses #232 

An example .platform.app.local.yaml file might be:
```yaml
hooks:
    build: |
        mv profiles/my_profile/my_file.txt my_file.txt
    deploy: |
        drush -y updb
        drush cc drush
        drush -y fra
```

and the output would look like:
```
Running local build hook(s)
Saving build archive...
Running local deploy hook(s)
No database updates required                                                                                        [success]
'all' cache was cleared.                                                                                            [success]
Finished performing updates.                                                                                        [ok]
'drush' cache was cleared.                                                                                          [success]
Current state already matches defaults, aborting.                                                                   [ok]
```

... you could even miss out the `-y` if you want to have confirmation prompts

### DIFF
--- a/src/Command/ProjectBuildCommand.php
+++ b/src/Command/ProjectBuildCommand.php
@@ -35,6 +35,11 @@ class ProjectBuildCommand extends PlatformCommand
                 null,
                 InputOption::VALUE_NONE,
                 'Do not create or use a build archive'
+            )->addOption(
+                'no-deploy-hooks',
+                null,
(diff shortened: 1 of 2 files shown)


pjcdawkins:
Closing in favour of trying to run normal Platform.sh build hooks (which will require implementing a docroot subdir)

PR #238, opened 2015-02-15
Accommodate apps without 'toolstacks' in build command.

pjcdawkins:


src/Local/Toolstack/Symfony.php
```diff
@@ -48,6 +29,7 @@ public function install()
         $symfonyResources = CLI_ROOT . '/resources/symfony';
 
         // Create and symlink configuration files.
+        // @todo are these actually useful?
```
pjcdawkins: So I'm going to need some help with answering this.

There may be other useful things we can do for Symfony users.

PR #255, opened 2015-03-17
Add sql and sql-dump commands

pjcdawkins:
Launch a MariaDB shell: `platform sql`

Execute a single statement: `platform sql "select * from users where name like 'alice'"`

Import local SQL file into the remote database: `platform sql < dump.sql`

Dump remote database to a local SQL file: `platform sql-dump`


pjcdawkins:
- [ ] get connection details from PLATFORM_RELATIONSHIPS

PR #261, opened 2015-03-25
Git command

pjcdawkins:
Reopening #256


pjcdawkins:
This would be `local:git` if #265 is merged

pjcdawkins:
#277 is a more flexible approach

PR #267, opened 2015-03-26
Specify default --source-dump for Drush sql-sync

pjcdawkins:
Helps with Drush 7


pjcdawkins:
Not practical at the moment as the --source-dump command is different between Drush 6 and 7

PR #275, opened 2015-04-13
Allow inherited variables to be disabled

pjcdawkins:
Pending support in the API


pjcdawkins:
- [ ] remove BC when all projects are refreshed
 - [ ] add a way to re-enable a disabled variable

PR #284, opened 2015-05-09
Use the new $PLATFORM_DOCUMENT_ROOT env var

pjcdawkins:


pjcdawkins:
- needs all projects to be refreshed with platform-git 1.13.4-1

PR #297, opened 2015-05-25
platform url should get the HTTPS URL of an environment

pjcdawkins:
This adds an heuristic to look through the environment's routes to find an HTTPS URL.

... or perhaps this should wait until there is a way to specify a default route.


OriPekelman:
see https://github.com/platformsh/platform-git/pull/78 and https://github.com/platformsh/platform-toolstack/pull/12

pjcdawkins:
We need a way to find the default URL but this is better than nothing

Issue #299, opened 2015-05-28
Allow building without symlinks


pjcdawkins:
Helps with Vagrant, and possibly other local workflows.

It would also help with doing #243 for any type of project.

pjcdawkins:
fixed in `dev-development`

PR #302, opened 2015-06-03
Allow building without symlinks

pjcdawkins:
Addressing #299


pjcdawkins:
There are still two symlinking instances:
1. Symlinking files from 'shared' to 'sites/default' in Drupal builds
2. Symlinking the whole build to 'www'

Not sure whether those should be eliminated too.

PR #310, opened 2015-06-22
[WIP] Add 'platform check' command

pjcdawkins:


pjcdawkins:
might need to rename this as it could be confused with 'checkout'

pjcdawkins:
Until we have a comprehensive way to create/write/check config files, 'platform build' is just as useful as this. It will throw a clear exception on invalid YAML.

PR #317, opened 2015-07-05
Automatically wait for environment activation when needed

pjcdawkins:
Ensures that these commands will succeed to find a remote SSH / public URL, if they ever can:

* `platform drush`
* `platform ssh`
* `platform relationships`
* `platform sql`
* `platform sql-dump`
* `platform url`

Might need to be broader in scope


pjcdawkins:
Preferred strategy:

1. check if the environment has an SSH URL
2. if not, reload and check again (once)
3. if it still doesn't have an SSH URL, return an error, suggesting why it might be (building or inactive)

PR #320, opened 2015-07-10
Support new config format for type/flavor

pjcdawkins:
- [x] needs tests

### DIFF
--- a/src/Local/LocalBuild.php
+++ b/src/Local/LocalBuild.php
@@ -1,6 +1,7 @@
 <?php
 namespace Platformsh\Cli\Local;
 
+use Platformsh\Cli\Exception\InvalidConfigException;
 use Platformsh\Cli\Helper\FilesystemHelper;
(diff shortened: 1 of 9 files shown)


src/Local/LocalBuild.php
```diff
@@ -180,7 +180,11 @@ public function getAppConfig($appRoot)
     public function getToolstack($appRoot, array $appConfig = array())
     {
         $toolstackChoice = false;
-        if (isset($appConfig['toolstack'])) {
+        // Forward compatibility with new toolstack config format.
+        if (isset($appConfig['type']) && isset($appConfig['build']['flavor'])) {
+            $toolstackChoice = $appConfig['type'] . ':' . $appConfig['build']['flavor'];
+        }
```
damz: `build.flavor` is actually optional, you can consider it to be `composer` (a new alias for `symfony`) if it is not specified.

PR #337, opened 2015-09-01
Rename 'backup' to 'snapshot'

pjcdawkins:
`platform environment:backup` -> `platform snapshot:create`
`platform environment:backup --list` -> `platform snapshots`
`platform environment:restore` -> `platform snapshot:restore`


pjcdawkins:
@oripekelman @kazanir any opinions on this sort of BC break? it may be possible to create a sort of hidden alias for the previous command name

OriPekelman:
If possible please use the aliases if not I would change nothing. 

Consistency is important.. but snapshot/backup are accepted synonyms.

From our docs: "The backup creates a complete snapshot of the environment."

pjcdawkins:
Well then we shouldn't have changed the name in the first place :) OK I'll look into the alias thing, if not I'll close this PR.

PR #338, opened 2015-09-06
New installer instructions

pjcdawkins:
Stop recommending Composer, as it causes problems with dependency conflicts.


pjcdawkins:
Todos:
 - [ ] It seems the Composer installer for Windows doesn't actually make git or php available (?)

Probably not fixable:
* MINGW64 appears to not allow `| php`
* MINGW64 doesn't allow login, Symfony Console can't cope with the hidden input:

  ```
  Your password: 'C:\Users\INNOMI~1\AppData\Local\Temp/hiddeninput.exe' is not recognized as an internal or external command,
operable program or batch file.
  ```

PR #347, opened 2015-09-27
Form library

pjcdawkins:
Experimental

Advantages:
* unifies command options with interactively entered values; potential for interactivity on many commands with minimal code
* removes symfony/options-resolver dependency
* relatively lightweight and simple compared with symfony/form and symfony/options-resolver
* values are normalized *before* they are validated, unlike options-resolver

Todo:
- [x] compare again with symfony-console-form library
- [x] compare with other CLI approaches (e.g. drupal console)
- [x] write tests
- [x] document
- [x] test performance impact (all forms are configured on application initialize)
- [x] split into separate library

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,12 +4,12 @@
     "license": "MIT",
     "require": {
         "doctrine/cache": "~1.4.2",
+        "platformsh/console-form": "0.0.3",
         "platformsh/client": "0.1.25",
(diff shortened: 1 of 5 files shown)


ralt:
Maybe this would be better as a separate library?

pjcdawkins:
Absolutely, just needs all the above first

PR #353, opened 2015-10-17
View a specific relationship property

pjcdawkins:
e.g.
```sh
platform relationships --property database.0.username
```

c.f. https://github.com/platformsh/platformsh-docs/pull/198


oxyc:
Even better :+1:

Issue #370, opened 2015-11-06
`drush @pid.branch ssh` should have the --cd equal the .platform.app.yaml web:document_root


joshmiller83:
When using platform to run drupal inside of a folder (as is the case for sites originally hosted on Acquia) an you try to ssh into the environment using the site alias, it wrongly passes "/app/public" and the ssh connection is refused. If we are passing a "cd" then it should equal the one defined in the .platform.app.yaml as the document root.

![image](https://cloud.githubusercontent.com/assets/1440242/11005817/2658fe60-848e-11e5-8169-217cef2efacf.png)

pjcdawkins:
The trouble is, that as well as the code complexity of checking your local .platform.app.yaml files for the document_root of each application, your local repository isn't a guarantee of what the remote environment contains - guessing that the document_root matches up on both is also flawed.

But this could still be implemented.

pjcdawkins:
The PR really does show the complexity ("10 changed files with 343 additions and 147 deletions") and it's still no guarantee that the remote root will be correct, but it might be good enough

damz:
On the other hand, it doesn't make sense to generate the remote aliases from the local app config, no? The config can and will be different depending on the environment.

pjcdawkins:
That could be the user's problem for now, until we have an API to find the real config for each app (per env)

damz:
@pjcdawkins In that case that sounds like a big change with little gain, no? Or is the architecture reusable later?

pjcdawkins:
Yes it's a big change, although the code gets a little bit refactored so the additions are not all bad. The gain depends on how long it would last as a stop-gap, before we implement an apps API. The code is not exactly reusable but yes the general ideas are there (and all that remains is to swap the information source, from local config to remote API).

pjcdawkins:
Improved in v2.7.0. I'm going to amalgamate various multi-app related issues into a new follow-up.

Issue #342, opened 2015-09-15
Drush aliases broken in multi-app projects


jurgenhaas:
When updating the drush aliases with ```platform drush-aliases``` and if the project contains more than one app, then the aliases are brokwn.

For the local one, the doc-root should point to [PROJECT]/www/[APP-NAME]

For the remote one, the remote-user should be [PROJECT-ID]-master--[APP-NAME]

Any idea how we could preserve those changes until the pltform CLI is fixing this?

pjcdawkins:
I don't think there is a workaround right now beyond copying the generated alias file and giving it another name.

One thing I could work on is how to get the doc root correct for the local alias.

For the remote alias, it's not possible to adapt to multi-app until we have implemented an API so that the CLI knows enough about what apps are on what remote environment (see #159, #326).

pjcdawkins:
Improved in v2.7.0, although I notice there may still be an issue with the local alias.

jurgenhaas:
This is great, I've just tested this and can confirm that the remote-user is set correctly as well as the root directory for the local alias. Excellent work!

@pjcdawkins what issue with the local alias are you referring to?

pjcdawkins:
Ah, well thanks for testing! To be honest I forgot that I'd implemented the local alias fix too... I've just checked and it's there.

Issue #268, opened 2015-03-27
Drupal toolstack hard-coded document root


petebarnett:
In a Platform.sh application, the document root can be set in the app config (web.document_root)

For a Drupal application, the build process needs to copy "settings.php" and create some symlinks in "sites/default" within the document root.
The Drupal toolstack however assumes that "sites/default" always exists in the root of the repository, even though this is configurable, and therefore changeable.

I'd like to propose that the CLI should honour the app config, and determine the path to "sites/default" based on this where necessary.

I have implemented the feature locally, and so would be happy to make a pull request after further testing.

pjcdawkins:
Agreed. We should take the opportunity to give builds a proper subdirectory structure and run the same build hooks as on Platform. It's going to make "vanilla" sites more complex though - what should happen there?

IIRC the configured document root is available in LocalBuild::getAppConfig().

BTW if you are making a pull request, note that we are now on version 2.0.0-alpha2 - I reckon there is still the possibility of BC breaking changes, if necessary, before 2.0.0.

pjcdawkins:
I've made a bit of a start in #269, I'm not sure it's the right direction so I'd be interested to see what you did

pjcdawkins:
Committed - https://github.com/platformsh/platformsh-cli/commit/3d1a515e1fb58170339bb18f0e8d139e5b4f297f

Issue #159, opened 2014-11-15
Support multiple apps per project


OriPekelman:
And is it documented?

OriPekelman:
Multiple public urls will be left for the moment for the developer to configure .. right ? We don't know what webserver is used. 

We can try and help with a magic nginx conf (and a *.local.platform.sh domain that will always resolve to 127.0.0.1 ) that will set the root dependent on the  host (with project/env/app taken into account) etc.. 

Better probably to leave this to the developer. Its a bit hard to do on our site, and quite easy on hers.

As for drush aliases: Dunno.

pjcdawkins:
Improved in v2.7.0. I'm going to amalgamate various multi-app related issues into a new follow-up.

PR #371, opened 2015-11-06
Use local app structure to find multi-app information

pjcdawkins:
Queries the local application structure, if possible, to make many commands aware of multiple applications. Not a perfect strategy, but a good stop-gap until there is an API to find the same information.

Includes `platform ssh`, `platform relationships`, `platform drush`, `platform sql`, `platform sql-dump`, and `platform drush-aliases`.

see #370 and #268 and #342 and #159


pjcdawkins:
Now also makes multi-app SSH much easier:

screen shot 2015-11-10 at 20 54 43

PR #383, opened 2015-11-25
Add `p` alias

pjcdawkins:


pjcdawkins:
better left to the docs :innocent:

PR #389, opened 2015-12-02
Add app commands to list apps and read config

pjcdawkins:


pjcdawkins:
Get the type of the app in a given project/environment (if 1 app):

```
platform app:config -p p43m25zns6k2c -e master --property type
```

Issue #395, opened 2015-12-13
RunOtherCommand and --quiet, output still enabled


vincenzo:
As part of my daily job I am developing an extended version of the CLI with many additional commands, so I use `RunOtherCommand` quite a lot. Most times I am ok with keeping the output, but sometimes I'd like to suppress it. I had expected that passing `--quiet` (see example below) to the command would do this, but it doesn't. Should it? 

```php
      $this->runOtherCommand('get', [
        '--quiet' => InputOption::VALUE_NONE,
        '--yes' => InputOption::VALUE_NONE,
        '--environment' => $testEnv,
        '--no-build' => InputOption::VALUE_NONE,
        'id' => $input->getOption('project'),
        'directory' => $siteDir,
      ]);
```

pjcdawkins:
Ah I thought those flags should be set to `true`, not `VALUE_NONE`.

By the way I wasn't expecting anyone to be extending this so I haven't bothered about internal backwards compatibility; the only BC is in commands/options/arguments

vincenzo:
I can't try now, but are you saying that if I used true it should work? Because VALUE_NONE works for the rest of them (essentially, it's a 1).

pjcdawkins:
It's done here:
https://github.com/symfony/console/blob/master/Application.php#L752
and yes `InputOption::VALUE_NONE` is `1`, not identical to `true`. So I think `true` should work

vincenzo:
Ah. I see now. I couldn't browse the code before. I'd better change all similar instances with true instead of what I had been using. Thanks Patrick.

vincenzo:
Hi, @pjcdawkins.
While you're right about using `true`, the issue seems to be another. 
Passing `--quiet` to `runOtherCommand()` won't have any effect because the `configureIO()` is invoked within `Symfony\Component\Console\Application::run()` before the `doRun()` method is invoked (which is the one that actually invokes the commands). So, if `platform` was executed without `--quiet`, that would stick throughout all the execution flow, even with the commands run within other commands using `runOtherCommand()`.

What I am doing at the moment to work around this is to change the verbosity of the output according to my needs, e.g. 

```php
      // Save original verbosity.
      $originalVerbosity = $output->getVerbosity();
      // Suppress output.
      $output->setVerbosity(OutputInterface::VERBOSITY_QUIET);
      $this->runOtherCommand('branch', [
        '--quiet' => TRUE,
        '--force' => TRUE,
        '--project' => $pid,
[...]

PR #398, opened 2015-12-14
Ability to pass --quiet onto runOtherCommand().

vincenzo:
This is related to https://github.com/platformsh/platformsh-cli/issues/395.

Passing `--quiet` to `runOtherCommand()` won't have any effect because the `configureIO()` is invoked within `Symfony\Component\Console\Application::run()` before the `doRun()` method is invoked (which is the one that actually invokes the commands). So, if `platform` was executed without `--quiet`, that would stick throughout all the execution flow, even with the commands run by other commands using `runOtherCommand()`.

This is an attempt to provide a solution for this, if appropriate.

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -1242,7 +1242,18 @@ abstract class CommandBase extends Command implements CanHideInListInterface
 
         $this->debug('Running command: ' . $name);
 
-        return $command->run($cmdInput, $this->output);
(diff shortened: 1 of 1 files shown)


pjcdawkins:
Here's a suggestion: #399, and then you could do:
```
runOtherCommand('get', $arguments, new NullOutput())
```

vincenzo:
Closing this in favour of #399.

PR #399, opened 2015-12-14
Allow an $output to be passed to runOtherCommand()

pjcdawkins:


pjcdawkins:
This removes the `$input` argument because this can be persisted as a property of the command (I'm not sure why I never did that). Having an `$output` argument lets you, for example, pass the Symfony `NullOutput()` class, which suppresses all output.

vincenzo:
Makes sense. Did you test this? I could quickly. 
I only ask because when I tried a similar approach (temporarily replacing the output for a single command), the output never came back for the rest of the process (which seems absurd, so I probably messed something up when I tried...).

pjcdawkins:
Yes, I tested this briefly

vincenzo:
Well, in that case, it's good. I'll close mine :)

PR #400, opened 2015-12-14
Remove automatic build from 'platform get'

pjcdawkins:


vincenzo:
+1 for this - I always use `get` with `--no-build`

PR #401, opened 2015-12-14
SSH tunnels

pjcdawkins:
Successor to #86

### DIFF
--- a/src/Command/Tunnel/TunnelOpenCommand.php
+++ b/src/Command/Tunnel/TunnelOpenCommand.php
@@ -1,0 +1,147 @@
+<?php
+namespace Platformsh\Cli\Command\Tunnel;
+
+use Platformsh\Cli\Util\ProcessManager;
+use Platformsh\Cli\Util\RelationshipsUtil;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class TunnelOpenCommand extends TunnelCommandBase
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('tunnel:open')
+            ->setDescription("Open SSH tunnels to an app's relationships");
+        $this->addProjectOption();
+        $this->addEnvironmentOption();
+        $this->addAppOption();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->validateInput($input);
+        $project = $this->getSelectedProject();
+        $environment = $this->getSelectedEnvironment();
+
+        if ($environment->id === 'master') {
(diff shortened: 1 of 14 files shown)


src/Command/Tunnel/TunnelOpenCommand.php
```diff
@@ -0,0 +1,198 @@
...
+
+                try {
+                    $processes[$pidFile] = $this->createTunnelProcess($sshUrl, $remoteHost, $remotePort, $localPort);
+                    $processes[$pidFile]->start(function ($type, $buffer) use ($log) {
+                        $log->writeln($buffer);
+                    });
+                }
+                catch (\Exception $e) {
+                    $this->stdErr->writeln(sprintf('Failed to open tunnel: %s', $e->getMessage()));
+                    unset($processes[$pidFile]);
```
ralt: Is this fine if `$processes[$pidFile]` doesn't exist?

pjcdawkins: Ah, no, I think it would trigger an undefined index notice

src/Command/Tunnel/TunnelOpenCommand.php
```diff
@@ -0,0 +1,198 @@
...
+
+                $start = isset($localPort) ? $localPort + 1 : 30000;
+                $localPort = PortUtil::getPort($start);
+                $tunnelInfo = [
+                    'projectId' => $project->id,
+                    'environmentId' => $environment->id,
+                    'appName' => $appName,
+                    'relationship' => $name,
+                    'key' => $key,
+                    'remotePort' => $remotePort,
```
ralt: It looks like `remotePort`, `remoteHost` and `localPort` aren't used?

pjcdawkins: I'm going to write all that info to a file, so you can have a `tunnel:status` command or similar to show all the tunnels you opened

ralt: Ah, definitely makes sense.

src/Command/Tunnel/TunnelOpenCommand.php
```diff
@@ -0,0 +1,199 @@
...
+
+        ProcessUtil::killParent($error);
+
+        if (!count($processes)) {
+            return $error ? 1 : 0;
+        }
+
+        while (count($processes)) {
+            sleep(1);
+            foreach (array_keys($processes) as $pidFile) {
```
ralt: Isn't it possible to have `$error === true` at this point? It seems to be completely ignored afterwards.

src/Util/PortUtil.php
```diff
@@ -0,0 +1,126 @@
+<?php
+
+namespace Platformsh\Cli\Util;
+
+class PortUtil
+{
+    protected static $unsafePorts = [
```
ralt: Should it be a `const`?

pjcdawkins: PHP <5.6 constants have to be scalar

Issue #406, opened 2015-12-30
[WIP] 3.x.x roadmap


pjcdawkins:
Major BC changes
* Use Guzzle 6 (drops PHP 5.4 support)
* Change file structure to use the main repository
  * move higher level files like '.platform-project', 'builds', 'shared', and '.build-archives' to .platform/local inside the repo, with a helper to add this to .gitignore
  * change project ID detection to use the Git remote, in place of or in addition to .platform-project
  * no longer store old uncompressed builds, just the current and the previous one (storing old builds never seemed to have a purpose)
  * remove 'platform clean'
  * remove 'platform init' (or perhaps this would just add the Git remote)
  * add migration tool to move from legacy file structure

Minor BC changes
* #393 
* #388 

Non-BC changes
* Symfony 3 compatibility
* Use Symfony Console Style Guide (http://symfony.com/blog/new-in-symfony-2-8-console-style-guide)

PR #410, opened 2016-01-09
3.x new file structure, etc.

pjcdawkins:
See #406 

- [x] detect project just by git remote, and make directories dynamically, so no 'init' required
- [x] add a little backwards compatibility esp. for scripted use
- [x] some kind of simple migration step from any command
- [ ] identify code that can be removed
- [ ] refactor


pjcdawkins:
Moved to 3.x branch.

PR #444, opened 2016-05-13
Display last update in 'platform environments'

mrkschan:


pjcdawkins:
As discussed this is a bit confusing: there are at least 3 types of 'last update time' - when the environment resource itself was updated, when the last activity finished, and when the code was last updated/changed. The `activity:list` command gives more info for the latter 2 types.

PR #445, opened 2016-05-18
Improve CliConfig coverage by forced config reset

pawpy:
No source changes; only test changes to improve the CliConfig class coverage by forced config reset.


pjcdawkins:
Makes sense, thanks

PR #449, opened 2016-05-23
Display create time in 'platform users'

mrkschan:


pjcdawkins:
- Yes we should display that date - thanks for the PR.
- [ ] Maybe `Date added` would be clearer? We are showing when the user was added to the project/environment, not when the user account was created.
- [ ] This depends on #450 (still in progress, a bit)
- [ ] We shouldn't create a `formatDateTime()` method every time we need to display a date - that's a detail
- The web UI doesn't show this much detail, by the way - should it?

pjcdawkins:
Unfortunately the `created_at` property appears to be inaccurate in the API: as far as I can tell, it changes very frequently, perhaps when the user information is synced.

PR #450, opened 2016-05-23
Adapt tables to the terminal width

pjcdawkins:


pjcdawkins:
Pros:
* Takes only 1-2ms per table
* seems to work

Cons:
* 211 lines of code
* Not scientific, just MVP with minimal research..
* Should really be contributed upstream (to Symfony Console)

PR #451, opened 2016-05-26
Adapt activity commands to show project activities

pjcdawkins:
FAO @mrkschan


pjcdawkins:
@Kazanir has pointed out we should not have this magic `-e 0` thing to show the project activities - it's really ugly.

So, I should add a new option to allow the user to specify something like `--all-environments` (or `--level=project`)...

PR #459, opened 2016-06-10
Sort projects and environments

pjcdawkins:


pjcdawkins:
... really, this should happen in the API, so you can combine it with `count` and filters

PR #461, opened 2016-06-12
Load all data for each project before caching

pjcdawkins:


pjcdawkins:
Too slow when there are too many projects

PR #467, opened 2016-06-20
[WIP] Add app:init command

pjcdawkins:
- [ ] Allow init from examples, as a shortcut
- [ ] Services/relationships
- [ ] Web/locations configuration
- [ ] Mounts
- [ ] Build flavors
- [ ] Disk size(s)
- [ ] More complicated things: hooks, dependencies, extensions, crons

PR #481, opened 2016-07-14
Add app:init command

Crell:
This PR replaces #467, as that one is from a 3rd party repo so I couldn't push to it.

Adds an "app:init" command, which will add viable if generic .platform\* files to a repository.

Current status: The command runs and produces files, enough to allow it to be pushed by Git.  The `routing.yaml` file is functional. The `.platform.app.yaml` file is not. The `services.yaml` file has some examples commented out, with instructions and a link to the documentation.
- [ ] Allow init from examples, as a shortcut
- [x] Routing
- [ ] Services/relationships
- [ ] Web/locations configuration
- [ ] Mounts
- [ ] Build flavors
- [ ] Disk size(s)
- [ ] More complicated things: hooks, dependencies, extensions, crons


src/Command/App/AppInitCommand.php
```diff
@@ -0,0 +1,244 @@
...
+"http://{default}/":
+  type: upstream
+  upstream: "{$applicationName}:http"
+"http://www.{default}/":
+  type: redirect
+  to: "http://{default}/"
+
+END;
+
+        (new Filesystem())->dumpFile($projectRoot . '/.platform/routing.yaml', $routingYaml);
```
mrkschan: Is this new? shouldn't it be .platform/routes.yaml?

Crell: D'Oh. This is what peer review is for. :smile: (routing.yml is what Drupal calls its files, so that's what's burned into my brain.)

Crell:
@pjcdawkins Thought: Should we be treating each language/version as a target, or should we be treating each stack as a target?  Damien is right that a wizard to build up the full, say, Drupal 8 configuration is going to be way way too complicated.  But, instead of options "php 5.5", "php 5.6", "nodejs" etc. we could offer "Generic PHP", "Drupal 7', "Drupal 8", "Symfony 3", "Generic NodeJs", etc.  Those files would be FAR more consistent, and the customization rather trivial (just relationships, basically).  

Some customization is still likely appropriate by the user, but far less than if we left out, say, the entire rules section.  It would also allow us to almost entirely eliminate the "generic $language" example repositories, or else make them simply an artifact of this script.

Does that seem reasonable?

PR #498, opened 2016-08-17
Add domain:update and domain:get commands

pjcdawkins:


mrkschan:
Is SSL cert being shown?

pjcdawkins:
@mrkschan yes, in the `platform domain:get` command. The certificate, not the private key (of course)

PR #502, opened 2016-08-30
Non-interactive login command.

slashrsm:
Works only with 2FA disabled.

### DIFF
--- a/src/Command/Auth/LoginNonInteractiveCommand.php
+++ b/src/Command/Auth/LoginNonInteractiveCommand.php
@@ -1,0 +1,67 @@
+<?php
+namespace Platformsh\Cli\Command\Auth;
+
+use GuzzleHttp\Exception\BadResponseException;
+use Platformsh\Cli\Command\CommandBase;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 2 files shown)


src/Command/Auth/LoginNonInteractiveCommand.php
```diff
@@ -0,0 +1,66 @@
...
+
+    protected function configure()
+    {
+        $this
+            ->setName('nilogin')
+            ->setDescription('Non-interactive log in to ' . self::$config->get('service.name'));
+        $help = 'Use this command to log in to your ' . self::$config->get('service.name') . ' account in a non-interactive way.'
+            . "\n\nYou can create an account at:\n    <info>" . self::$config->get('service.accounts_url') . '</info>'
+            . "\n\nIf you have an account, but you do not already have a password, you can set one here:\n    <info>" . self::$config->get('service.accounts_url') . '/user/password</info>';
+        $this->setHelp($help);
```
ralt: Maybe add that TFA must be disabled?

slashrsm: Done.

pjcdawkins:
This would mean entering a password on the command-line, which would remain in shell history, etc., in plain text. Maybe that's fine on an otherwise secure environment, but that's not something we can assume.

There is "API token" support for this reason (see https://github.com/platformsh/platformsh-cli#customization).

slashrsm:
Thanks! I guess this makes it PR useless.

Issue #503, opened 2016-08-31
is there a way to prevent backing up local builds?


hanoii:
Looking at the code I didn't find a one.

pjcdawkins:
Would you like one? 😃  See the pull request linked above.

hanoii:
And regardless of the follow up on the PR, awesome, thanks!

pjcdawkins:
Added in [v3.5.0](https://github.com/platformsh/platformsh-cli/releases/v3.5.0)

PR #504, opened 2016-09-05
Add --no-backup option to build command

pjcdawkins:
As an answer to #503

### DIFF
--- a/src/Command/Local/LocalBuildCommand.php
+++ b/src/Command/Local/LocalBuildCommand.php
@@ -57,6 +57,12 @@ class LocalBuildCommand extends CommandBase
                 'Do not create or use a build archive'
             )
             ->addOption(
+                'no-backup',
+                null,
+                InputOption::VALUE_NONE,
+                'Do not back up the previous build'
+            )
+            ->addOption(
                 'no-cache',
                 null,
                 InputOption::VALUE_NONE,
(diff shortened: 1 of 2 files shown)


hanoii:
Can this option be set globally, in the config file, so I don't have to alias it or pass it to every command line?

pjcdawkins:
Hm. Is there a problem with the backups - do they fail, or take too long?

hanoii:
Hmm, no, not really, I just don't wan't them/need them taking up space locally :). Why would I need a build backup if the whole thing is git driven..  It might be a nice addition to be able to set command line arguments globally on a config file, kind of how drush works, but this is just widening the scope to a more general PR, just sharing the thought while I had it.

ralt:
A one-off would probably be a bad idea, but a global config file a la .gitconfig could be a nice idea to have per-command defaults.

pjcdawkins:
OK, let's call that another issue. It would add a bit of complexity, and I'm not clear whether it should be global or per-project for example.

@hanoii the built code is backed up because it's not git-driven _after_ it's built - there is a chance that a user has made some changes to a third-party library, for example. Or the user changed compiled CSS directly, etc... So the backup gives them an extra chance to save that work.

hanoii:
@pjcdawkins ok, I think that's a fair point. I do think that is more the exception than the norm, and the backups will likely/mostly not be necessary, so it might be a nice idea to follow up the global/per project setting, and if this can be accepted, the better. I always push to have an option that to have none.

Thanks!

PR #505, opened 2016-09-08
Add 'title' filter to project:list

vincenzo:
Currently, the `project:list (projects)` command allows filtering by region (host). However, if your account holds hundreds of sites like mine, it becomes helpful to have a filter on the project title. This is how I have implemented it in my custom version of the CLI: very simply, the filter will show all projects that have a certain (sub)string in their title. The match is case insensitive by default and not configurable.

Thoughts?

### DIFF
--- a/src/Command/Project/ProjectListCommand.php
+++ b/src/Command/Project/ProjectListCommand.php
@@ -20,7 +20,7 @@ class ProjectListCommand extends CommandBase
             ->setDescription('Get a list of all active projects')
             ->addOption('pipe', null, InputOption::VALUE_NONE, 'Output a simple list of project IDs')
             ->addOption('host', null, InputOption::VALUE_OPTIONAL, 'Filter by region hostname')
-            ->addOption('title', null, InputOption::VALUE_REQUIRED, 'Filter by title')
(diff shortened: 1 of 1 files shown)


vincenzo:
@floretan Fair enough on the optional. 

Re the filter in the API, perhaps you're right. But at the moment even the website has a search implemented (would seem) on the site itself.

pjcdawkins:
Actually both `title` and `host` should have the `InputOption::VALUE_REQUIRED` flag. It means that you have to specify the _value_ for the option if you use it. The options themselves remain optional (that's why they are called options 😃 ).

vincenzo:
@pjcdawkins Shall I revert my last commit, then?

vincenzo:
@pjcdawkins I restored `VALUE_REQUIRED` for the new `title` option, but also changed it for the `host` option. Let me know.

pjcdawkins:
Fine with me - thanks! I squash-merged to make the history simpler.

pjcdawkins:
Released in [v3.5.0](https://github.com/platformsh/platformsh-cli/releases/v3.5.0)

PR #512, opened 2016-09-19
Allowing to pass on additional dump commands to sql-dump

hanoii:
Mostly so that you can pass on addition commands, like skipping tables, or outputting only schema, etc.


pjcdawkins:
Thanks for this.

In 8d1fd698006ac74a273a5bbcc5fef08c10c1c35b I've added three explicit options (`--table`, `--exclude-table`, and `--schema-only`) which work with both postgresql and mysql.

I think having those options is going to help with backwards compatibility for any future changes: for example, the `extra` argument could cause problems should the command be switched from mysqldump to some other utility.

PR #519, opened 2016-10-07
Option for not deleting remote branch

hanoii:
I think removing the remote branch by default when removing is dangerous. It's not often, but we are enforcing a feature-branching model, and it's possible that we would want some features to be kept remotely without an environment.

Also I noticed that although there are --yes, --no and --delete-branch, we don't really have a way to not delete it from the command line. So added that option as well.

### DIFF
--- a/src/Command/Environment/EnvironmentDeleteCommand.php
+++ b/src/Command/Environment/EnvironmentDeleteCommand.php
@@ -191,7 +191,7 @@ EOF
                 if ($questionHelper->confirm("Are you sure you want to delete the environment <comment>$environmentId</comment>?")) {
                     $deactivate[$environmentId] = $environment;
                     if (!$input->getOption('no-delete-branch')) {
-                        if ($input->getOption('delete-branch') || ($input->isInteractive() && $questionHelper->confirm("Delete the remote Git branch too?", false))) {
+                        if ($input->getOption('delete-branch') || ($input->isInteractive() && $questionHelper->confirm("Delete the remote Git branch too?"))) {
                             $delete[$environmentId] = $environment;
                         }
                     }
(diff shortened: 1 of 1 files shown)


hanoii:
Now that I re-thought this, I think more often than not, you would remove the remote branch, but still worth having a command line option to make sure you don't.

Not sure what your thoughts are on the default, feel free to change it.

pjcdawkins:
The command doesn't really "remove the branch by default" right now.

It currently requires **either** `$input->isInteractive()` **or** `--delete-branch`, i.e.:

```sh
# Deactivate the environment (do not delete the branch):
platform environment:delete --environment xyz --yes
# Deactivate the environment and delete the branch:
platform environment:delete --environment xyz --delete-branch --yes
# Prompt to deactivate the environment and prompt to delete the branch:
platform environment:delete --environment xyz
# Prompt to deactivate the environment and then (without prompt) delete the branch:
platform environment:delete --environment xyz --delete-branch
```

So I guess you're looking for an option to see the prompts about the environment(s), but never to delete the branch. Sounds reasonable.

```sh
# Prompt to deactivate the environment (do not delete the branch):
platform environment:delete --environment xyz --no-delete-branch
```

It might need something to handle `--no-delete-branch --inactive`, i.e. "please delete all inactive environments, but don't delete them"... which should be treated as an invalid combination of options, IMO.

hanoii:
By "remove by default" I meant that the prompt to remove the remote branch defaults to Y.

I was actually looking for this:

```
# Deactivate the environment (do not delete the branch):
platform environment:delete --environment xyz --yes
```

I didn't realize that `--yes` was both disabling interaction and by such, not removing the branch. My immediate assumption, as per the help:

```
  -y, --yes                      Answer "yes" to all prompts; disable interaction
```

was that it answered all to all prompts, and as such, would both remove the environment AND the remote branch. 

I guess though that adding the no-delete-branch adds even more flexibility.

PR #520, opened 2016-10-07
environment:branch to track upstream

hanoii:
I noticed that when you branch with platform, your local branch is not tracking. This adds the remote tracking.


pjcdawkins:
Thanks. I suspect this can be done with a bit less code (I wonder, do we need to `git fetch`?): https://github.com/platformsh/platformsh-cli/compare/development...pjcdawkins:branch-track-upstream-2

hanoii:
You do need `git fetch` because the branch is created both remotely and locally but because it was never **fetched**, setting the upstream will fail.

See https://github.com/platformsh/platformsh-cli/blob/development/src/Command/Environment/EnvironmentCheckoutCommand.php#L69-L83 which does a very similar thing. It's been a while but I remember it not working without the fetch.

PR #529, opened 2016-10-29
Permit a command to continue after updating

pjcdawkins:
- [x] find if there is a less fragile way to autoload needed classes after update
- [x] find if there is a less fragile way to get the executable name
- [x] consider UX (is "Continue? [Y/n]" clear enough)


ralt:
For the UX, I'd prefer something like `Run <command> with the new version?`... can't comment on the autoloading.

Thanks for that either way!

PR #532, opened 2016-11-08
Create CLI command for quick-read on SQL disk usage.

bbujisic:
The `EnvironmentSqlSizeCommand` class is a little bit dirty, especially as it does not return the same data for MySQL and PostgreSQL.

I guessed two types of reports were sufficiant for the time being: Overview and Details. Detailed reports can be requested with `--details` option.

Out of the box, MySQL allows XML responses, so I added a `--xml` option for that. 

Review would be more than appreciated.

Thanks,
Branislav

### DIFF
--- a/src/Command/Environment/EnvironmentSqlSizeCommand.php
+++ b/src/Command/Environment/EnvironmentSqlSizeCommand.php
@@ -7,7 +7,6 @@ use Platformsh\Cli\Helper\ShellHelper;
 use Platformsh\Cli\Util\RelationshipsUtil;
 use Platformsh\Cli\Util\Table;
 use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Output\BufferedOutput;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Yaml\Yaml;
 
@@ -32,8 +31,6 @@ class EnvironmentSqlSizeCommand extends CommandBase {
 
         /** @var ShellHelper $shellHelper */
         $shellHelper = $this->getHelper('shell');
-        $bufferedOutput = new BufferedOutput();
-        $shellHelper->setOutput($bufferedOutput);
 
         // Get and parse app config.
         $args = ['ssh', $sshUrl, 'echo $' . self::$config->get('service.env_prefix') . 'APPLICATION'];
--- a/src/Command/Environment/EnvironmentSqlSizeCommand.php
+++ b/src/Command/Environment/EnvironmentSqlSizeCommand.php
@@ -1,0 +1,160 @@
+<?php
+namespace Platformsh\Cli\Command\Environment;
+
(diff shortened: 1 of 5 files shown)


src/Command/Environment/EnvironmentSqlSizeCommand.php
```diff
@@ -0,0 +1,185 @@
...
+        $services = Yaml::parse(file_get_contents($projectRoot . '/.platform/services.yaml'));
+        $allocatedDisk = $services[$dbServiceName]['disk'];
+
+        $this->stdErr->write('Querying database <comment>' . $dbServiceName . '</comment> to estimate disk usage. ');
+        $this->stdErr->writeln('This might take a while.');
+
+        /** @var ShellHelper $shellHelper */
+        $shellHelper = $this->getHelper('shell');
+        $command = ['ssh'];
+        if ($output->getVerbosity() >= OutputInterface::VERBOSITY_DEBUG) {
```
pjcdawkins: Adding these options doesn't have any effect, because the command is executed without printing output. These lines (relating to verbosity) should probably just be removed.

src/Command/Environment/EnvironmentSqlSizeCommand.php
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+        if (!$dbServiceName) {
+            $this->stdErr->writeln('Service name not found for relationship: ' . $database['_relationship_name']);
+            return 1;
+        }
+
+        // Load services yaml.
+        $services = Yaml::parse(file_get_contents($projectRoot . '/.platform/services.yaml'));
+        $allocatedDisk = $services[$dbServiceName]['disk'];
```
pjcdawkins: This should fail (with `return 1` and a message, or `a RuntimeException`, I don't mind) if the `$services[$dbServiceName]['disk']` is empty. That will save time and potentially save you a division by zero later on.

src/Command/Environment/EnvironmentSqlSizeCommand.php
```diff
@@ -70,6 +70,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
         // Load services yaml.
         $services = Yaml::parse(file_get_contents($projectRoot . '/.platform/services.yaml'));
         $allocatedDisk = $services[$dbServiceName]['disk'];
+        if(empty($allocatedDisk)) {
```
pjcdawkins: To be fussy, I meant
```php
if (!empty($services[$dbServiceName]['disk'])) {
  $allocatedDisk = $services[$dbServiceName]['disk'];
} else {
  $this->stdErr->writeln('The allocated disk size could not be determined for service: ' . $dbServiceName);
  return 1;
}
```

bbujisic: Arrr, I really am not concentrated today!

PR #545, opened 2016-11-24
Add unix shell as requirement

mrkschan:
Unless PowerShell is supported, we better let user know we need linux/unix shell.


pjcdawkins:
I think for Windows users, "Linux/Unix" would be confusing. We'd need to list what actual applications would work and which are recommended, e.g.

* On Windows: [Bash on Ubuntu on Windows](https://msdn.microsoft.com/en-gb/commandline/wsl/about) (recommended), or a Bash-compatible shell such as [Git Bash](https://git-for-windows.github.io/), Cygwin, or MinGW.

mrkschan:
Updated, please merge and squash this PR with the github button :p

PR #551, opened 2016-12-07
Document setting environment variables.

bbujisic:
I had troubles finding documentation for custom environment variables (https://docs.platform.sh/development/environment-variables.html#custom-environment-variables), so I thought adding an example with `env:` prefix would make sense.


src/Command/Project/Variable/ProjectVariableSetCommand.php
```diff
@@ -30,6 +30,7 @@ protected function configure()
         $this->addExample('Set the variable "example" to the Boolean TRUE', 'example --json true');
         $this->addExample('Set the variable "example" to a list of values', 'example --json \'["value1", "value2"]\'');
         $this->addExample('Set the variable "example" to the string "abc", but only at build time', 'example abc --no-visible-runtime');
+        $this->addExample('Set the variable "EXAMPLE" outside of the PLATFORM_VARIABLES to the string "abc" by prefixing the variable name with "env:"', 'env:EXAMPLE abc');
```
bbujisic: I just stole "You can create simple environment variables outside of the PLATFORM_VARIABLES value by prefixing the variable name with env:." sentence from https://docs.platform.sh/development/environment-variables.html

PR #552, opened 2016-12-07
Add push command, new SSH arguments, and DI

pjcdawkins:
Intended to address https://github.com/platformsh/platformsh-docs/pull/418

Needs some refactoring.

The `--identity-file` option (and the resulting temporary GIT_SSH file stuff) should be provided for other SSH/Git commands.


src/Command/Environment/EnvironmentPushCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+        foreach (['force', 'dry-run'] as $option) {
+            if ($input->getOption($option)) {
+                $command .= ' --' . $option;
+            }
+        }
+
+        $sshOptions = [];
+        if ($input->getOption('no-wait')) {
+            $sshOptions[] = "-o 'SendEnv PLATFORMSH_PUSH_*'";
```
ralt: We should either push `PLATFORMSH_*` all the time, or only push `PLATFORMSH_PUSH_NO_WAIT` in this case, no?

src/Command/Environment/EnvironmentPushCommand.php
```diff
@@ -0,0 +1,111 @@
...
+            $command = 'PLATFORMSH_PUSH_NO_WAIT=1 ' . $command;
+        }
+        if ($identityFile = $input->getOption('identity-file')) {
+            if (!file_exists($identityFile)) {
+                $this->stdErr->writeln('File not found: ' . $identityFile);
+                return 1;
+            }
+            $sshOptions[] = '-i ' . escapeshellarg($identityFile);
```
ralt: I'd use the long option there :-)

src/CliConfig.php
```diff
@@ -71,6 +71,21 @@ public function getUserConfigDir()
...
 
     /**
+     * @param bool $create
+     *
+     * @return string
+     */
+    public function getTempDir($create = true)
+    {
+        $dir = $this->getUserConfigDir() . '/tmp';
+        if (!is_dir($dir) && $create && !mkdir($dir, 0700, true)) {
```
ralt: I would `!@is_dir($dir)` to avoid any warning. which would seem to appear in case the folder doesn't exist? Same for `!@mkdir` in case of permission issue.

src/Command/Environment/EnvironmentPushCommand.php
```diff
@@ -0,0 +1,111 @@
...
+        $tempFile = null;
+        if (!empty($sshOptions)) {
+            $sshCommand = 'ssh ' . implode(' ', $sshOptions) . ' $*';
+            $tempFile = $this->writeSshFile($sshCommand);
+            $command = 'GIT_SSH=' . escapeshellarg($tempFile) . ' ' . $command;
+        }
+
+        $result = $shellHelper->executeSimple($command, $projectRoot);
+
+        if ($tempFile) {
```
ralt: I'd do that in a `try/finally` in case `executeSimple` throws.

src/Command/Environment/EnvironmentPushCommand.php
```diff
@@ -0,0 +1,111 @@
...
+            if (!file_exists($identityFile)) {
+                $this->stdErr->writeln('File not found: ' . $identityFile);
+                return 1;
+            }
+            $sshOptions[] = '-i ' . escapeshellarg($identityFile);
+        }
+
+        $tempFile = null;
+        if (!empty($sshOptions)) {
+            $sshCommand = 'ssh ' . implode(' ', $sshOptions) . ' $*';
```
ralt: Small thing: I'd add a newline at the end... for esthetic purposes. Yes, it's a temporary file. I have a problem.

src/Command/Environment/EnvironmentPushCommand.php
pjcdawkins: I should make it `PLATFORMSH_PUSH_NO_WAIT` then.

`PLATFORMSH_*` would be overly inclusive, as there are potentially `PLATFORMSH_CLI_*` vars that are intended only for local.

PR #562, opened 2017-01-17
Clarify that self-update output messages are about platform.sh CLI

fuzzbomb:
Platform.sh CLI occasionally checks for updates, even when `platform self:update` is not the command being run.

If the CLI is called from a custom script, then it isn't always clear that the message is about updating the platform.sh CLI tool.

Example:
We use [Docksal](http://docksal.io/) `fin` custom commands for for tasks such as dumping a DB from a platform.sh environment, and importing it into the local Docksal database container. It's a little bit confusing to type a `fin db-download master` command and see a message about checking for updates - is it updating `fin` or updating `platform`?

This PR adds the name "platform.sh CLI" to some of the SelfUpdate output messages:

 - At the start of the output ("This instance of the platform.sh CLI ..." and "Checking for platform.sh CLI updates..."), and
 - At the end of the output, if the update actually took place ("Successfully updated platform.sh CLI...").

All these messages are still less than 75 characters, so will likely fit on one line.

### DIFF
--- a/src/Service/SelfUpdater.php
+++ b/src/Service/SelfUpdater.php
@@ -86,8 +86,12 @@ class SelfUpdater
     {
         $currentVersion = $currentVersion ?: $this->config->get('application.version');
         $manifestUrl = $manifestUrl ?: $this->config->get('application.manifest_url');
+        $applicationName = $this->config->get('application.name');
         if (!extension_loaded('Phar') || !($localPhar = \Phar::running(false))) {
-            $this->stdErr->writeln('This instance of the platform.sh CLI was not installed as a Phar archive.');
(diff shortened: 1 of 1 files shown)


pjcdawkins:
Thanks. The name of the application is actually [configurable](https://github.com/platformsh/platformsh-cli/blob/0fde438001a107bd5cf864edce9eed08c37a0b37/config.yaml#L3) so I've committed an update to the PR to reflect that.

You can disable automatic updates, by the way, in `~/.platformsh/config.yaml` (see https://github.com/platformsh/platformsh-cli/#customization).

fuzzbomb:
Thanks @pjcdawkins

PR #575, opened 2017-03-05
WIP: Add config:generate commands

pjcdawkins:
Renamed and restructured from #481 

- [ ] download templates from URLs (i.e. override `getTemplate()` in subcommands)
- [ ] improve `config:generate` UX (i.e. listing the subcommands in a friendlier way)
- [ ] add a template engine for more complicated options


OriPekelman:
We should create templates for Ruby, Python, Golang, Java (they can be extremely minimal)

PR #591, opened 2017-04-25
Rely on date_default_timezone_get() instead of UTC fallback

Pierstoval:
PHP should be aware of the current timezone, so fallback to UTC would be implicit by using `date_default_timezone_get()`, because if `date.timezone` is not set in `php.ini`, default one is UTC.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -243,7 +243,8 @@ class Application extends ParentApplication
      */
     protected function setDefaultTimezone()
     {
-        $timezone = 'UTC';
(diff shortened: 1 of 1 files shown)


pjcdawkins:
That sounds sensible, thanks for the PR. Does this fix a specific/practical issue?

Pierstoval:
This doesn't fix any issue, actually, but allows falling back to PHP config instead of hard-coded timezone. For Windows + WSL for example, it fixes the issue for me, because in WSL, the different links in `/etc/` do not exist.

PR #592, opened 2017-04-25
Ssh issue with windows

### LINKED ISSUE #590: Windows SSH cannot resolve host
When using Windows, SSH doesn't work because it seems that the performed ssh command is not formated correctly:

```batch
e:\dev\www\app> platform ssh -e master -A myapp
DEBUG Finding the project root
DEBUG Project root found: E:\dev\www\app
DEBUG Project {project_id} is mapped to the current directory
DEBUG Validated input
Running command: ssh '-o' 'SendEnv TERM' "{project_id}-master-a1b2c3d4--myapp@ssh.eu.platform.sh"
ssh: Could not resolve hostname '-o': Hô
```

When I switch to WSL & Bash, it works perfectly and the request has the same format except the double-quotes are converted to single-quotes.

Actually, the solution would simply be to replace all quotes with double-quotes, because when I change the command manually with double quotes, it works.

Pierstoval:
This fixes #590 , hope this does not introduce regressions with other usages of the `getSshArgs()` method 😕 

My very first contrib on this project, so I might miss something. 

I let you guys and CI tell me where I'm wrong 😄 👍

### DIFF
--- a/src/Command/Environment/EnvironmentSshCommand.php
+++ b/src/Command/Environment/EnvironmentSshCommand.php
@@ -20,7 +20,7 @@ class EnvironmentSshCommand extends CommandBase
             ->setName('environment:ssh')
             ->setAliases(['ssh'])
             ->addArgument('cmd', InputArgument::OPTIONAL, 'A command to run on the environment.')
-            ->addOption('pipe', null, InputOption::VALUE_NONE, "Output the SSH URL only.")
+            ->addOption('pipe', null, InputOption::VALUE_NONE, 'Output the SSH URL only.')
             ->setDescription('SSH to the current environment');
         $this->addProjectOption()
              ->addEnvironmentOption()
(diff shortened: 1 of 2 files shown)


src/Command/Environment/EnvironmentSshCommand.php
```diff
@@ -20,7 +20,7 @@ protected function configure()
             ->setName('environment:ssh')
             ->setAliases(['ssh'])
             ->addArgument('cmd', InputArgument::OPTIONAL, 'A command to run on the environment.')
-            ->addOption('pipe', null, InputOption::VALUE_NONE, "Output the SSH URL only.")
+            ->addOption('pipe', null, InputOption::VALUE_NONE, 'Output the SSH URL only.')
```
Pierstoval: Yeah, just because there's no need for double quotes here, sorry for diff annoyance 🤣

Pierstoval:
Thanks, can't wait for new release 👍

PR #600, opened 2017-05-23
Add automatic configurtion for Solr cores.

MacWeber:
It gets the default config and also custom multi core configurations.


pjcdawkins:
This should be part of https://github.com/platformsh/template-builder first

PR #611, opened 2017-07-26
Add redeploy command

pjcdawkins:


pjcdawkins:
Closing on the grounds that this shouldn't ever be needed. Equivalent command:
```sh
platform vset _redeploy "$(date)"
```

PR #625, opened 2017-09-05
Close pipes before closing proc

luxifer:
As stated in http://php.net/proc_open all pipes must be closed, before closing the process to prevent any memory leaks.


damz:
Given the descriptor spec, that uses only existing file descriptors, `$pipes` would be always empty, no?

luxifer:
https://3v4l.org/Ur4nt You're right, my bad.

PR #633, opened 2017-09-25
mount:list, mount:upload and mount:download commands

markushausammann:
... to easily look up mounts and rsync files up or down. I've never worked with any of the used components before so I'm pretty sure some stuff (like verbose error reporting, etc.) can be improved but I think it's a pretty solid base for you guys to work with. push and pull duplicate quite some code too and I think you might want to refactor part of this into some service or similar.

### DIFF
--- a/src/Command/Mount/MountPullCommand.php
+++ b/src/Command/Mount/MountPullCommand.php
@@ -1,110 +1,86 @@
 <?php
+
 namespace Platformsh\Cli\Command\Mount;
 
-use Platformsh\Cli\Command\CommandBase;
-use Platformsh\Cli\Local\LocalApplication;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
-class MountPullCommand extends CommandBase
+class MountPullCommand extends MountSyncCommandBase
 {
 
+    /**
+     * {@inheritdoc}
+     */
     protected function configure()
     {
         $this
             ->setName('mount:pull')
-	        ->setAliases(['mpull'])
-            ->setDescription('Download the contents of mounts')
-	        ->addOption('refresh', null, InputOption::VALUE_NONE, 'Whether to refresh the cache');
-	    $this->addProjectOption();
-	    $this->addEnvironmentOption();
-	    $this->addAppOption();
+            ->setAliases(['mpull'])
+            ->setDescription('Download the contents of a mount')
+            ->addOption('mount', null, InputOption::VALUE_REQUIRED, 'The mount (as an app-relative path)')
+            ->addOption('target', null, InputOption::VALUE_REQUIRED, 'The directory to which files will be downloaded');
+        $this->addProjectOption();
+        $this->addEnvironmentOption();
+        $this->addAppOption();
     }
 
-	/**
-	 * @param InputInterface  $input
-	 * @param OutputInterface $output
-	 *
-	 * @return int|null|void
-	 * @throws \RuntimeException
-	 * @throws \Platformsh\Client\Exception\OperationUnavailableException
-	 * @throws \Platformsh\Client\Exception\EnvironmentStateException
-	 */
-	protected function execute(InputInterface $input, OutputInterface $output)
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
     {
         $this->validateInput($input);
 
-	    /** @var \Platformsh\Cli\Service\RemoteEnvVars $envVarService */
-	    $envVarService = $this->getService('remote_env_vars');
+        $sshUrl = $this->getSelectedEnvironment()
+            ->getSshUrl($this->selectApp($input));
 
-	    $sshUrl = $this->getSelectedEnvironment()
-		    ->getSshUrl($this->selectApp($input));
+        $appConfig = $this->getAppConfig($sshUrl);
 
-	    $result = $envVarService->getEnvVar('APPLICATION', $sshUrl, $input->getOption('refresh'));
-	    $appConfig = json_decode(base64_decode($result), true);
+        if (empty($appConfig['mounts'])) {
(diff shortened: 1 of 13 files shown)


markushausammann:
the first commit message should correctly be `add mount:list (mounts) command ` ...

pjcdawkins:
Some notes, for you or me:

* It would be nice if this worked with or without a local clone, i.e. base the target path on `$this->getProjectRoot() ?: getcwd()` instead of using `LocalApplication`.
* I'd like to be able to specify the local target path (for `mount:pull`) and local source path (for `mount:push`), e.g. with a `--target` option. It could have a default as the path to the currently built (via `platform build`) local app's mount paths, which would end up as something like `.platform/local/shared/files`.
* These should be usable non-interactively, so there should be an argument/option for specifying which mount is to be pulled/pushed (it would be specified by its full path, I guess, e.g. `platform mount:push --source ./files --mount /app/public/sites/default/files`).

src/Command/Mount/MountPullCommand.php
```diff
@@ -0,0 +1,110 @@
...
+	 *
+	 * @param $path
+	 * @param $sshUrl
+	 * @param $appRoot
+	 *
+	 * @return bool
+	 */
+	private function runSync($path, $sshUrl, $appRoot)
+	{
+		$command = sprintf('rsync -az %s:/app%s/ %s%s', $sshUrl, $path, $appRoot, $path);
```
Crell: Possibly include a --verbose flag on the command, which will add a `v` flag to the rsync command?  I often find that very useful as an informal progress meter.

Crell:
This will be super-useful, thank you!  

Do we want to include an "all" flag to sync all defined mounts at once, or are we concerned about that pulling too much?  (Eg, it would include a temp directory or the drush backup directory on a Drupal site, which is generally not needed.)

markushausammann:
It's already very useful for us and I'd appreciate if you (we) don't try to implement the Mercedes right now but release the Fiat and then iterate to make it better. I agree with the suggested additions but I don't need them. What I implemented is basically the 'local CMS dev' use case.

markushausammann:
> Do we want to include an "all" flag to sync all defined mounts at once, or are we concerned about that pulling too much?

I first added the all chice at 0 but then realized that I can't foresee a situation where I would really want to pull (let alone push) all folders at once. Like cache, tmp, logs, etc. as you mentioned.

markushausammann:
One additional thing: I was wondering about the `--iconv=utf-8-mac,utf-8` flag in case of uploads from Mac. I don't actually understand why that's needed and if it's not a relic from the past, but if it really is needed, it should probably be included as a flag/option as well.

pjcdawkins:
I implemented my comments (mostly to pass time while travelling), adding the options, a default, and non-interactive use. @markushausammann I don't want to get in the way - could you try it out to see whether my assumptions have disrupted your workflow?

markushausammann:
@pjcdawkins yes, you actually did disrupt my workflow... the default target directory should be the mapped directory in the root of the project as I implemented it because that will in 80% of the cases be the target directory and I don't want to enter that.

markushausammann:
It's very easy to fix. Just suggest the mapped path as default and let me hit enter.

markushausammann:
Btw. there is something wrong with the 'directory' question, it ends with `::`. And I think there should be some explanation regarding what kind of path people should enter, relative to what? Project root, absolute, ...?

markushausammann:
Right now, the expected path is not a logical expectation in case of multi-app projects. I chose an app first, then I chose a mount, then I need to enter a path where I have to take a mental step back to the project root instead of the app root. So here's how I think it should be implemented:

- when asking for path, show mapped path as default option `[mapped path]:`
- when asking for path in multi-app project, ask for relative paths to app root (so relative to `.platform.app.yaml`)
- when wrong path is entered, add explanation to error message that a path relative to app root is requried

pjcdawkins:
The above keeps the assumption that local relative paths are relative to the CWD / current working directory (whether you are in the app root, project root, or elsewhere).

markushausammann:
This is great!

pjcdawkins:
I have been wondering about a hard problem: naming things... IMHO, the terms "push" and "pull" may be confusing as they sound like version control operations.

Naming options (in my own order of preference):

1. Use `mount:up-sync` and `mount:down-sync` (could automatically be [shortened](https://symfony.com/doc/current/components/console/usage.html#shortcut-syntax) to `m:u` and `m:d`)
2. Use `mount:sync-up` and `mount:sync-down` (shorten badly, to `m:sync-u` and `m:sync-d`)
3. Use `mount:sync:up` and `mount:sync:down` (shorten to `m:s:u` and `m:s:d`)
4. Use `files:up-sync`, `files:down-sync` (shorten to `f:u` and `f:d`)
5. Use `file:sync:up` and `file:sync:down` (shorten to `f:s:u` and `f:s:d`)
6. Keep the status quo, `mount:push`, `mount:pull`
7. Use `rsync:up` and `rsync:down`

I think `mount:list (mounts)` can stay the same regardless.

markushausammann:
I think it should definitely be `mount:` as it is in one group with `mount:list` and should stay grouped that way. It also has a high self-documentary value. As to push and pull I followed rsync naming (https://linux.die.net/man/1/rsync) for consistency but it's true that it can be confused with version control operations. On the basis of that argumentation, why not simplify it to `mount:up` and `mount:down`? Would need any shorthands then either. I didn't want to use the "sync" word because it's not really a sync (sync is understood as bi-directional), it's just an intelligent up- or download.

pjcdawkins:
Ah, I didn't realise the rsync docs used push and pull.

I think I like `mount:up` and `mount:down` 🤔 

It could be `mount:upload` and `mount:download`, and thanks to the built-in shorthand, those would automatically work as `mount:up` and `mount:down` anyway. Likewise for up-sync/down-sync, or even up-rsync/down-rsync.

markushausammann:
Yeah, `mount:upload` and `mount:download` are awesome!

pjcdawkins:
Thanks @markushausammann, this is a really great feature!

markushausammann:
Well, thank you!

markushausammann:
Will you tag a release soon?

PR #641, opened 2017-11-10
WIP: Use new platformsh/client and platformsh/oauth2

pjcdawkins:
Depends on https://github.com/platformsh/platformsh-client-php/pull/21

Would require a new major version (because the PHP minimum version increases from 5.5.9 to 5.6.0).


pjcdawkins:
#699

PR #654, opened 2017-11-29
Use Git data API

pjcdawkins:
- [ ] Ensure version 2.7 is available on all projects first
- [x] Naming?... `read-file` vs `cat-file`, etc.
- [x] Caching?... will cache size grow out of control? (since it can't be nested against the project/environment)
- [x] Handling binary files? (e.g. skip caching?)


pjcdawkins:
Merged for release on approx. 6th December

PR #675, opened 2018-01-24
Remove timezone code

pjcdawkins:
It overrides the user's existing PHP configuration in favour of the system timezone, and it is probably unnecessary if Symfony Console doesn't have it upstream.


pjcdawkins:
Composer does this:
https://github.com/composer/composer/blob/847ebed6e6999022f08ca615d69c5664f41d866b/src/Composer/Console/Application.php#L70-L72

pjcdawkins:
Superseded by #680

Issue #677, opened 2018-01-29
User defined site aliases are prefixed with '@' when generated with drush-aliases


GROwen:
When running `platform drush-aliases` it's expected that any existing user defined aliases are preserved. This is the case however they are altered and prefixed with an `@` symbol.

Any user defined aliases that exist in `{alias_group}.site.yml` are prefixed with an `@` symbol on each run of `platform drush-aliases` resulting in aliases like `@alias_group.@userdefined` on the first run and `@alias_group.@@userdefined`.

pjcdawkins:
Fixed in v3.28.0

PR #678, opened 2018-01-29
Fixed user defined aliases being prefixed with '@'

GROwen:
Addresses #677


pjcdawkins:
Thanks very much for finding this and the PR - it fixes the issue but I think the real source of the bug is in `DrushAlias::getExistingAliases()`, so a slightly clearer fix would be #679.

GROwen:
Closing pull request in favour of #679

PR #682, opened 2018-02-03
Compress and remove whitespace from the generated Phar

pjcdawkins:
Produces 783KB archive - the current one is 3.3MB. It would possibly have an adverse effect on performance or compatibility.

### DIFF
--- a/box.json
+++ b/box.json
@@ -30,5 +30,10 @@
   ],
   "main": "bin/platform",
   "output": "platform.phar",
-  "stub": "stub.php"
+  "stub": "stub.php",
+  "compression": "BZ2",
+  "compactors": [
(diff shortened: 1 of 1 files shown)


box.json
```diff
@@ -30,5 +30,10 @@
   ],
   "main": "bin/platform",
   "output": "platform.phar",
-  "stub": "stub.php"
+  "stub": "stub.php",
+  "compression": "BZ2",
```
damz: You might want `zlib` instead. Bzip2-compressed .ZIP archives are... rare.

pjcdawkins:
The compactors alone make no difference to the size (?), and the GZ compression adds ~50ms to each command run on my machine.

PR #683, opened 2018-02-03
[db:dump] Remove --no-autocommit and simplify mysqldump args

pjcdawkins:
* `--no-autocommit` causes problems with importing large tables into Galera
* `--opt` and `--quote-names` are enabled by default, so we do not need to specify them
* `--single-transaction` recommends `--quick`, but that is also enabled by `--opt`, which is default


damz:
:+1:

damz:
@pjcdawkins Have you opened an issue against Drush?

PR #688, opened 2018-02-21
Add redeploy command, and some automatic --no-wait detection

pjcdawkins:
- [x] Should we have `--no-wait` by default? Currently I think that would be a confusing inconsistency (edit: see discussion below - this adds a `shouldWait()` method that detects cron/hooks)
- [ ] Wait for availability of platform.sh version 2.8.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "guzzlehttp/guzzle": "^5.3",
         "guzzlehttp/ringphp": "^1.1",
         "platformsh/console-form": ">=0.0.16 <2.0",
-        "platformsh/client": ">=0.15.0 <2.0",
+        "platformsh/client": ">=0.16.0 <2.0",
         "symfony/console": "^3.0 !=3.2.5 !=3.2.6",
         "symfony/yaml": "^3.0 || ^2.6",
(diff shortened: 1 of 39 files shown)


Crell:
Given the most likely use case at the moment would be from a cron hook to setup auto-redeploy, and not having `--no-wait` there would cause a deadlock, I think a strong case can be made for defaulting to `--no-wait` here.  Which suggests maybe we need a `--yes-wait` flag?

pjcdawkins:
Defaulting to `--no-wait` here (yes it would need an alternative `--wait`) means that it's not consistent with other commands where deadlocking is also a problem, like `snapshot:create` (`backup`) and `variable:set` (wherever that's being used).

The alternative which I'm considering is making some kind of `shouldWait()` heuristic, which attempts to detect if it's being run on a Platform.sh environment (under Dash with certain environment variables), and if so prints a warning and avoids waiting.

Crell:
If we can semi-reliably detect that the CLI is running from cron/deploy and always force that to no-wait, that would be the best option, I agree.  I'm totally on board with still including the bright flashing warnings in the documentation to manually include `--no-wait`.

PR #710, opened 2018-05-10
Command lazy-loading; service auto-configuration and autowiring

pjcdawkins:
* Shaves off 40-50ms from every command invocation (assuming the DI container is cached, and if using the non-aliased command name, for now).
* Allows eventual removal of `getService()` and auto-wired constructor injection.
* Allows auto-discovery of commands in the `src/Command/` directory, no need to register anymore (except perhaps for command aliases in future).


pjcdawkins:
To refactor CommandBase the components will have to be:

- [ ] auto updates check
- [ ] input validation, i.e. project and environment selection
- [ ] event subscriber (login required and environments updated) which is aware of other services
- [ ] utility for dispatching to other commands
- [ ] decoration for providing command examples and hidden aliases
- [ ] detection of container environment
- [ ] I/O handy methods like stdErr property and debug()

PR #722, opened 2018-05-31
Adds a scaffold command

xtfer:
Adds a basic scaffold command to create default Platform.sh files.

`scaffold:default (scaffold)`

This copies files from a scaffold folder located in the `resources` directory, where the path `resources/<template-name` corresponds to a usable scaffold.

Only the default scaffold exists currently, for PoC evaluation. The default is currently a PHP application with standard routes and a services file with services commented out.

If this works, additional scaffolds could be created for common runtimes and frameworks, potentially reducing the need for example repositories in some cases. Other possible behaviours include fetching the scaffold from a public git repository instead of storing it locally, which would allow users to build their own scaffolds.

### DIFF
--- a/resources/scaffold/default/.platform.app.yaml
+++ b/resources/scaffold/default/.platform.app.yaml
@@ -1,0 +1,42 @@
+# This file describes an application. You can have multiple applications
+# in the same project.
+
+# The name of this app. Must be unique within a project.
+name: app
+
+# The type of the application to build.
+type: php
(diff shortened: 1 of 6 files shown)


xtfer:
The history is screwy. Will Close and fix.

pjcdawkins:
Seen #575?

xtfer:
Er, no... let me look at that :)

PR #769, opened 2019-01-16
Drush alias fixes

pjcdawkins:
- Use relative path for `root` (to account for differing app roots)
- A globally installed Drush 9 needs the aliases to be in `groupname.site.yml` (rather than `groupname.alias.yml`) but this isn't being picked up when in a particular site that uses Drush 8.
- Drush 9 has moved unknown keys such as `platformsh-cli-auto-remove` under an `options` key.

Todo:
- [x] appears to work with Drush 8.1
- [x] check if this works with other versions (particularly Drush 9+)

### DIFF
--- a/src/Service/Drush.php
+++ b/src/Service/Drush.php
@@ -255,18 +255,6 @@ class Drush
     }
 
     /**
-     * @return string
-     */
-    protected function getAutoRemoveKey()
-    {
-        return preg_replace(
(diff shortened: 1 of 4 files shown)


Berdir:
Did not actually test the changes here but manually removed the '/app/' and can confirm that this worked on both professional and enterprise with Drush 8.

PR #773, opened 2019-01-28
Only show user project choice if they have multiple

robbie-thompson:
Users with only a single project shouldn't be prompted to select the project they want to run the command for every time. When a user runs a command and only has one project, they will intend for the command to run for that project.


pjcdawkins:
Sorry, I'm going to close this due to the rationale in https://github.com/platformsh/platformsh-cli/pull/778#issuecomment-505407903

PR #778, opened 2019-02-15
Auto-select a project/environment when there is only 1

pjcdawkins:
With credit to @robbie-thompson in #773 

- [ ] are there commands where it would be surprising/dangerous to auto-select the project?
- [ ] are there commands where it would be surprising/dangerous to auto-select the environment?


pjcdawkins:
The CLI already uses many methods to select a project or environment

* command-line option (-p or -e)
* environment variable (PLATFORM_PROJECT and PLATFORM_BRANCH)
* mapping in .platform/project.yaml
* Git remote / branch / upstream

This would add a further method ("has only 1 in API"), which adds complexity, and furthermore is not controllable locally.

PR #783, opened 2019-02-26
fix(mongo): database authentification

Kocal:
Using v3.40.2, we have the following issue when trying to run `platform service:mongo:export`:

selection_248

I don't know if there is a test for this because I made a really quick fix, but this should do the work.


pjcdawkins:
Thanks @Kocal !

pjcdawkins:
Released in version v3.40.4 (update with `platform up`)

Kocal:
Nice :+1:

Kocal:
(and it's working fine)

PR #784, opened 2019-02-27
WIP: Pf 972 init generalized

OriPekelman:
Perfect is the enemy of done.. and I really want this...

* Continued work ... add all supported runtimes. 
* Made a bit more minimalistic.
* Detect missing configuration and add some autorun fun.

Still missing how to get runtime_versions..


OriPekelman:
I think this one can be closed :)

Issue #774, opened 2019-02-12
Can not tunnel:open


salines:
```
platform tunnel:open
Enter a number to choose a project:
  [0] ****** (******)
  [1] ********* (******)
 > 1
```
 
exit without any message

pjcdawkins:
Hi, could you re-run the command with `-v`? It would show up any PHP errors.

cferthorney:
I'm having identical issues. Currently running `3.40.4` PHP 7.3.2 using standard Homebrew configuration:
```
brew info php@7.3
php: stable 7.3.2 (bottled)
General-purpose scripting language
https://secure.php.net/
/usr/local/Cellar/php/7.3.2 (521 files, 77MB) *
  Poured from bottle on 2019-02-08 at 10:59:54
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/php.rb
==> Dependencies
Build: httpd ✘, pkg-config ✔
Required: apr ✔, apr-util ✔, argon2 ✔, aspell ✔, autoconf ✔, curl-openssl ✔, freetds ✔, freetype ✔, gettext ✔, glib ✔, gmp ✔, icu4c ✔, jpeg ✔, libpng ✔, libpq ✔, libsodium ✔, libzip ✔, openldap ✔, openssl ✔, sqlite ✔, tidy-html5 ✔, unixodbc ✔, webp ✔
==> Caveats
To enable PHP in Apache add the following to httpd.conf and restart Apache:
    LoadModule php7_module /usr/local/opt/php/lib/httpd/modules/libphp7.so

    <FilesMatch \.php$>
        SetHandler application/x-httpd-php
    </FilesMatch>

Finally, check DirectoryIndex includes index.php
    DirectoryIndex index.php index.html

The php.ini and php-fpm.ini file can be found in:
    /usr/local/etc/php/7.3/

To have launchd start php now and restart at login:
  brew services start php
Or, if you don't want/need a background service you can just run:
  php-fpm
==> Analytics
install: 33,993 (30 days), 109,962 (90 days), 457,168 (365 days)
install_on_request: 30,608 (30 days), 99,245 (90 days), 412,373 (365 days)
build_error: 0 (30 days)
```

I have xdebug and imagick installed, but uninstalling them seems to make no difference to the results.


```
user@machine ~/Development/{client}/Drupal/{project}/docroot (master) $ platform -v tunnels
No tunnels found.
[...]

PR #786, opened 2019-03-11
Disable pcre.jit in tunnel:open command to work around PHP bug #77260

pjcdawkins:
PHP bug 77260: https://bugs.php.net/bug.php?id=77260

Somehow this ini_set() doesn't fix the issue if it's included significantly later (but still before the `fork()`).

That leads me to suspect the introduction of other PCRE regexes before this line may reintroduce this bug, so this is fragile.

Addresses issue #774

### DIFF
--- a/bin/platform
+++ b/bin/platform
@@ -6,6 +6,19 @@ if (version_compare(PHP_VERSION, '5.5.9', '<')) {
     exit(1);
 }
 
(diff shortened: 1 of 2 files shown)


pjcdawkins:
Even with this workaround the command still segfaults in some circumstances, e.g. when adding `-v` - I guess this is because it triggers PCRE to start up at an earlier point.

PR #792, opened 2019-04-02
Fleets

xtfer:
This adds new commands for managing a Fleet of projects, which can then be piped into the `multi` command.

A Fleet is currently a named key in a file `.platform/fleets.yaml`, which contains a `project` key containing an array of project IDs. This can be edited manually or by using the following CLI commands:

- `fleet:list`: List all fleets and their projects
- `fleet:add`: Add a new fleet
- `fleet:remove`: Remove a fleet
- `fleet:project-add`: Add a project ID to a fleet
- `fleet:project-remove`: Remove a project ID from a fleet

This PR combines work @Kazanir did on the multi command with a Fleet service class and commands.

### DIFF
--- a/src/Service/Filesystem.php
+++ b/src/Service/Filesystem.php
@@ -5,6 +5,7 @@ namespace Platformsh\Cli\Service;
 use Platformsh\Cli\Util\OsUtil;
 use Symfony\Component\Filesystem\Exception\IOException;
 use Symfony\Component\Filesystem\Filesystem as SymfonyFilesystem;
+use Symfony\Component\Yaml\Yaml;
 
 class Filesystem
 {
@@ -491,5 +492,37 @@ class Filesystem
             }
         }
         throw new \RuntimeException("Tar command not found");
+    }
+
+    /**
+     * @param $filename
+     * @return bool
+     */
+    public function fileExists($filename) {
+        return $this->fs->exists($filename);
+    }
+
+    /**
+     * @param string $filename
+     *
+     * @return array
+     */
+    public function readYamlFile($filename)
+    {
+        $contents = file_get_contents($filename);
+        if ($contents === false) {
+            throw new \RuntimeException('Failed to read file: ' . $filename);
+        }
+
+        return (array) Yaml::parse($contents);
+    }
+
+    /**
+     * @param $filename
+     * @param array $contents
+     */
+    public function createYamlFile($filename, array $contents) {
+        $yaml = Yaml::dump($contents);
+        $this->fs->dumpFile($filename, $yaml);
     }
 }
--- a/src/Service/Fleets.php
+++ b/src/Service/Fleets.php
@@ -2,7 +2,6 @@
 
 namespace Platformsh\Cli\Service;
 
-use Platformsh\Cli\Exception\RootNotFoundException;
 use Platformsh\Cli\Local\LocalProject;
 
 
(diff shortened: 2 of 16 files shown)


src/Service/Fleets.php
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * Get the name of the configuration file for fleets.
+     *
+     * @return string
+     */
+    protected function getConfigFileName()
+    {
+        return $this->config->get('service.project_config_dir') . '/fleets.yaml';
```
pjcdawkins: This will load from `.platform/fleets.yaml`, relative to whatever directory the CLI is running in. Problems:

* The `local.fleet_config` option added in `config.yaml` in this branch isn't being used.
* The `local.fleet_config` (being defined as .platform/local/fleets.yaml` would be excluded by Git by default, that's probably not wanted.
* This should probably be based on the project root directory, which could be passed in by the caller, or obtained through the `LocalProject` service's `getProjectRoot()` method.

src/Service/Filesystem.php
```diff
@@ -492,4 +493,36 @@ protected function getTarExecutable()
...
+    public function fileExists($filename) {
+        return $this->fs->exists($filename);
+    }
+
+    /**
+     * @param string $filename
+     *
+     * @return array
+     */
+    public function readYamlFile($filename)
```
pjcdawkins: The Symfony Yaml class already has `Yaml::parseFile()`

src/Service/Filesystem.php
```diff
@@ -492,4 +493,36 @@ protected function getTarExecutable()
...
+        }
+
+        return (array) Yaml::parse($contents);
+    }
+
+    /**
+     * @param $filename
+     * @param array $contents
+     */
+    public function createYamlFile($filename, array $contents) {
```
pjcdawkins: If readYamlFile() is removed (see above comment) I don't think this method is necessary either (then Filesystem needn't depend on Symfony Yaml).

src/Service/Fleets.php
```diff
@@ -0,0 +1,267 @@
...
+
+namespace Platformsh\Cli\Service;
+
+use Platformsh\Cli\Local\LocalProject;
+
+
+class Fleets
+{
+
+    const FLEET_DOES_NOT_EXIST = 0;
```
pjcdawkins: I'd prefer exceptions rather than error codes - getFleet() perhaps could return either a fleet object or false, but addFleet() and removeFleet() should just be relied on for their named purpose, not worrying about the file not existing etc.

(happy to rejig this myself)

src/Service/Fleets.php
xtfer: I think local.fleet_config is a remnant of Christian's work. I'll remove that.

It's entirely possible that you might want to run fleet commands without a project. For that reason I remove any LocalProject references.

src/Service/Fleets.php
xtfer: The case where a fleet does not exist is not an Exception, it's just incorrect input data, nor is it sufficiently complex to warrant the kind of cascade that Exceptions are useful for. 

So, we can do this, but it will be a lot more work for the same result. To make this functionally equivalent we'd need an Exception class for each case, then try/catch handling for those cases, which would just get us back to this point, since there is no further Exception handling required as we don't through any other kinds.

src/Service/Fleets.php
pjcdawkins: OK, but selecting `.platform/fleets.yaml` from anywhere you run the command is going to be problematic, the base path (the cwd) could be anything and would often be invisible to the user depending on their terminal/prompt.

The root path should be something well defined: either the local project root, and/or somewhere in the user's home directory, and/or we allow the user to override the path to the fleets.yaml.

src/Service/Fleets.php
pjcdawkins: IMO we wouldn't need a separate exception class or try/catch.. I'll propose something in a branch.

PR #804, opened 2019-06-04
Pass max_allowed_packet setting to mysqldump

danielcavanagh:
I was receiving this error while trying to dump by DB
```
> platform db:dump
Creating SQL dump file: dump.sql
mysqldump: Error 2020: Got packet bigger than 'max_allowed_packet' bytes when dumping table `cache_render` at row: 1024
```
max_allowed_packet is set in the DB service config but it isn't passed on to mysqldump. This patch passes the setting on and allows my dump to succeed


pjcdawkins:
That's odd, this configuration ends up in `my.cnf`

pjcdawkins:
.. the PR seems safe and helpful anyway, thanks @danielcavanagh !

PR #812, opened 2019-06-14
Adding more precise dbsize

matthiaz:
And some refactoring

### DIFF
--- a/src/Command/Db/DbSizeCommand.php
+++ b/src/Command/Db/DbSizeCommand.php
@@ -87,14 +87,14 @@ class DbSizeCommand extends CommandBase
             $propertyNames  = ['Allocated', 'Estimated Usage', 'Percentage Used'];
             $percentageUsed = $estimatedUsage * 100 / $allocatedDisk;
             $values = [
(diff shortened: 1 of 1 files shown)


pjcdawkins:
https://github.com/matthiaz/platformsh-cli/pull/1

pjcdawkins:
It seems the column `ALLOCATED_SIZE` doesn't exist in `information_schema.innodb_sys_tablespaces` for older MariaDB versions (works in 10.2, doesn't work in 10.0). We need to test for the existence of that column.

matthiaz:
> It seems the column `ALLOCATED_SIZE` doesn't exist in `information_schema.innodb_sys_tablespaces` for older MariaDB versions (works in 10.2, doesn't work in 10.0). We need to test for the existence of that column.

I fixed that.

matthiaz:
I also expanded on the 'cleanup' option
so you can do something like `platform db:size -p 3oupdxcaenytw -e master -r database1 --cleanup`

InnoDB does not automatically clean up deleted data, but running an ALTER TABLE Engine forces it to be recreated. 

--cleanup won't do this without first asking confirmation from the user.

PR #814, opened 2019-06-19
Build on platform.sh

pjcdawkins:
Don't know why I didn't think of this before...


pjcdawkins:
It built OK https://pr-814-4r3vpfa-mpaw4nkgdwaea.eu-2.platformsh.site/
but hasn't reported its status back to GitHub

PR #816, opened 2019-06-27
Show usage per database.

matthiaz:
Customer reported a bug in the new version. That caused him to see the same numbers for every database/relation. I had to add a new table to show the database sizes because we can't directly compare that against the allocated size. 
On the bright side: this does make it very clear where the storage usage is coming from.

image

### DIFF
--- a/src/Command/Db/DbSizeCommand.php
+++ b/src/Command/Db/DbSizeCommand.php
@@ -55,7 +55,6 @@ class DbSizeCommand extends CommandBase
         }
 
         if (!isset($database['service'])) {
-            $this->stdErr->writeln(print_r($database, true));
             $this->stdErr->writeln('Unable to find database service information.');
             return 1;
         }
@@ -269,13 +268,11 @@ class DbSizeCommand extends CommandBase
         $relationships = $this->getService('relationships');
         $connectionParams = $relationships->getDbCommandArgs('mysql', $database, '');
 
-        $return =sprintf(
+        return sprintf(
             "mysql %s --no-auto-rehash --raw --skip-column-names --execute '%s'",
             $connectionParams,
             $query
         );
-
-        return $return;
     }
 
     /**
--- a/src/Command/Db/DbSizeCommand.php
+++ b/src/Command/Db/DbSizeCommand.php
@@ -53,7 +53,9 @@ class DbSizeCommand extends CommandBase
             $this->stdErr->writeln('No database selected.');
             return 1;
         }
+
         if (!isset($database['service'])) {
+            $this->stdErr->writeln(print_r($database, true));
             $this->stdErr->writeln('Unable to find database service information.');
             return 1;
         }
@@ -67,9 +69,9 @@ class DbSizeCommand extends CommandBase
         $this->stdErr->writeln(sprintf('Checking database service <comment>%s</comment>...', $dbServiceName));
 
         $this->debug('Calculating estimated usage...');
-        $allocatedDisk = $service->disk * self::BYTE_TO_MEGABYTE;
+        $allocatedDisk  = $service->disk * self::BYTE_TO_MEGABYTE;
         $estimatedUsage = $this->getEstimatedUsage($sshUrl, $database);
-        $percentageUsed = round($estimatedUsage * 100 / $allocatedDisk);
+        $percentageUsed = round($estimatedUsage['__TOTAL__'] * 100 / $allocatedDisk);
 
         /** @var \Platformsh\Cli\Service\Table $table */
         $table = $this->getService('table');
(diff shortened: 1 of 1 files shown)


pjcdawkins:
> Customer reported a bug in the new version. That caused him to see the same numbers for every database/relation.
Sorry.. what's the bug?

matthiaz:
I removed the colouring on the database table, because we don't want to alarm the customer for no reason. There is nothing wrong with a customer having a single database ;-)

matthiaz:
Do note that this also fixes a UI bug that was introduced in the previous merge to development.

pjcdawkins:
Still scratching my head over what the UI bug is. Do you mean the per-relationship usage? That is/was misleading, as the disk exists independently of the relationships, so you'd need to see the full usage on the disk regardless of which relationship is used. Breaking down the usage per schema/table as well as showing the full usage is a useful feature.

matthiaz:
The problem is that you used to be able to specify which database you
wanted to see the size for.

We changed the way the size is calculated to be more accurate (file based
in stead of innodb based). And I forgot that you could specify a specific
database.

As a result, each database shows the total size.

I may have gone a little too far in fixing a bug by adding a feature that
changes the UI. But that was because I discovered that comparing the
allocated space to 1 database, is like comparing apples to oranges.

And that last one was indeed a problem in the previous version as well.

Hence: this is a feature, to solve a bug, which was discovered while fixing
another bug. :)

Sorry I made it a little confusing

matthiaz:
@pjcdawkins Do you want me to reclassify this as a feature request to merge this?

src/Command/Db/DbSizeCommand.php
```diff
@@ -78,19 +82,40 @@ protected function execute(InputInterface $input, OutputInterface $output) {
         $machineReadable = $table->formatIsMachineReadable();
         $showInBytes = $input->getOption('bytes') || $machineReadable;
 
-        $columns  = ['max' => 'Allocated disk', 'used' => 'Estimated usage', 'percent_used' => '% used'];
+        $columns  = ['max' => 'Allocated disk', 'used' => 'Estimated usage', 'percent_used' => 'Percentage used'];
```
pjcdawkins: There are some irrelevant changes in the diff, like this one

src/Command/Db/DbSizeCommand.php
```diff
@@ -78,19 +82,40 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
         $table->render([$values], $columns);
-
+        
         $this->showWarnings($percentageUsed);
+                
+
 
-        $this->showInaccessibleSchemas($service, $database);
+        $db_table   = $this->getService('table');
+        $db_columns = ['db' => 'Database', 'used' => 'Estimated usage', 'percent' => 'Percentage'];
```
pjcdawkins: Having two tables and naming the columns in both is going to be awkward, I believe it will break the `--columns` option.

src/Command/Db/DbSizeCommand.php
```diff
@@ -377,12 +418,13 @@ private function getMySqlUsage(HostInterface $host, array $database) {
      *
      * @return string
      */
-    private function formatPercentage($percentage, $machineReadable) {
+    private function formatPercentage($percentage, $machineReadable, $blnUseColour=true) {
```
pjcdawkins: Irrelevant / unnecessary change

src/Command/Db/DbSizeCommand.php
```diff
@@ -287,8 +318,18 @@ private function mysqlNonInnodbQuery(array $database, $excludeInnoDb = true)
...
-        $query = 'SELECT SUM(ALLOCATED_SIZE) FROM information_schema.innodb_sys_tablespaces;';
-
+        $deployment = $this->api()->getCurrentDeployment($this->getSelectedEnvironment());
+        $service    = $deployment->getService($database['service']);
+        $databases  = array_keys($service->configuration['endpoints'][$database['rel']]['privileges']);
+        
+        //convert database names to OR statements that the sys_tablespaces can use.
+        $strDbs     = implode(' OR ', array_map(
+                                        function($database) {
+                                            return "`NAME` LIKE \"$database/%\"";
```
pjcdawkins: It comes from the service configuration so it's probably OK, but it would be safer to escape the string for SQL here

src/Command/Db/DbSizeCommand.php
```diff
@@ -249,11 +277,12 @@ private function getMysqlCommand(array $database, $query) {
         $relationships = $this->getService('relationships');
         $connectionParams = $relationships->getDbCommandArgs('mysql', $database, '');
 
-        return sprintf(
+        $return =sprintf(
```
pjcdawkins: Unnecessary change

src/Command/Db/DbSizeCommand.php
```diff
@@ -99,6 +124,9 @@ protected function execute(InputInterface $input, OutputInterface $output) {
         return 0;
     }
 
+    private function showEstimatedUsage() {
```
pjcdawkins: Empty/unused function

src/Command/Db/DbSizeCommand.php
matthiaz: I can't use https://www.php.net/manual/en/mysqli.real-escape-string.php since we don't have a connection object.
Is addslashes sufficient?

src/Command/Db/DbSizeCommand.php
matthiaz: $blnUseColour : I use it, because we don't want to show colours when were simply listing the distribution of database usage size.
See: `platform db:size -p sudo27zgjpwqa -vvv -e WEBSUPPORT-170-office-renewal --relationship=management -D`

src/Command/Db/DbSizeCommand.php
matthiaz: changed

src/Command/Db/DbSizeCommand.php
matthiaz: removed

src/Command/Db/DbSizeCommand.php
matthiaz: I removed that change

src/Command/Db/DbSizeCommand.php
matthiaz: You're right. It does produce weird results. I moved that logic under a flag to make it optional.

src/Command/Db/DbSizeCommand.php
matthiaz: I added addslashes

PR #821, opened 2019-07-04
Allow `url` command to read from local environment

pjcdawkins:


src/Command/Environment/EnvironmentUrlCommand.php
```diff
@@ -14,26 +16,73 @@ protected function configure()
...
+        }
+        if (empty($routes)) {
             $output->writeln("No URLs found");
             return 1;
         }
 
+        $primaryUrl = false;
+        foreach ($routes as $route) {
+            if ($route->primary) {
+                $primaryUrl = $route->url;
```
pjcdawkins:
```suggestion
                $primaryUrl = $route->url;
                break;
```

src/Command/Environment/EnvironmentUrlCommand.php
```diff
@@ -14,26 +16,73 @@ protected function configure()
...
+        $urls = array_map(function (Route $route) {
+            return $route->url;
+        }, $routes);
+
+        // Sort URLs by preference (HTTPS first, shorter URLs first).
         usort($urls, [$this->api(), 'urlSort']);
 
+        // Shift the primary URL to the front.
+        if ($primaryUrl !== false) {
+            unset($urls[array_search($primaryUrl, $urls, true)]);
```
pjcdawkins: Maybe skip the reshuffle if the key is already 0

PR #828, opened 2019-07-31
Rename "Capacity" to "% Used" in mount:size command

pjcdawkins:
Also makes column machine names consistent with the db:size command


pjcdawkins:
From a suggestion by Jari Lana in the public Slack channel https://platformsh.slack.com/archives/C0JHEUHQD/p1564569371437500

PR #830, opened 2019-08-06
WIP: Overriding the ConsoleForm form class resolveOptions method

shawnawsu:
`project:create` will now take three different flags: 

- `--template` (requires a url to be passed)
- `--catalog` (will add a field to the interactive form)
- `--initialize` (will initialize the project after it's been created, this needs a few tweaks)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -80,6 +80,10 @@ service:
     - standard
     - medium
     - large
+  catalog:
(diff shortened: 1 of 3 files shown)


shawnawsu:
We're moving forward with #831 so this can be closed.

PR #831, opened 2019-08-06
Adds a secondary form to project:create for template, catalog, and initialize options

shawnawsu:
A new form will be built and served if the `--template` flag is passed. The fields are `template_url` which allows for the passing of a url, `catalog_url` which will list all options from the catalog, and `initialize` which will initialize the project after creation if selected.

### DIFF
--- a/src/Command/Project/ProjectCreateCommand.php
+++ b/src/Command/Project/ProjectCreateCommand.php
@@ -30,7 +30,7 @@ class ProjectCreateCommand extends CommandBase
 
         $this->addOption('check-timeout', null, InputOption::VALUE_REQUIRED, 'The API timeout while checking the project status', 30)
             ->addOption('timeout', null, InputOption::VALUE_REQUIRED, 'The total timeout for all API checks (0 to disable the timeout)', 900)
-            ->addOption('template', null, InputOption::VALUE_NONE, 'Choose a template on which to build your project.')
+            ->addOption('template', null, InputOption::VALUE_OPTIONAL, 'Provide a template url or choose a template from the provided list on which to build your project.', false)
             ->addOption('initialize', null, InputOption::VALUE_NONE, 'Initialize the project after it has been created.')
             ->addOption('current-repo', null, InputOption::VALUE_NONE, 'Automatically set remote after creation.')
             ->addOption('region', null, InputOption::VALUE_REQUIRED, 'The region to assign the project to.');
@@ -58,7 +58,14 @@ EOF
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $this->form = Form::fromArray($this->getFields($input->getOption('template')));
+        // Process the template field to determine if it's a url or just the flag.
+        //@todo How do we want to handle bad urls?
+        $catalog = false;
+        if ($input->getOption('template') !== false && empty(parse_url($input->getOption('template'), PHP_URL_PATH))) {
+            $catalog = true;
+        }
+
+        $this->form = Form::fromArray($this->getFields($catalog));
         $this->form->configureInputDefinition($this->getDefinition());
 
         /** @var \Platformsh\Cli\Service\QuestionHelper $questionHelper */
@@ -86,7 +93,10 @@ EOF
         }
         
         // Handle if no template is chosen look for a default one for the user.
-        if (empty($options['catalog'])) {
+        if ($catalog === false && !empty(parse_url($input->getOption('template'), PHP_URL_PATH))) {
+            $options['catalog'] = $input->getOption('template');
+        }
+        else {
             $options['catalog'] == NULL;
         }
 
--- a/src/Command/Project/ProjectCreateCommand.php
+++ b/src/Command/Project/ProjectCreateCommand.php
@@ -300,41 +300,6 @@ EOF
             'questionLine' => '',
             'default' => 'Untitled Project',
           ]);
-
-        //   $fields['template_option'] = new OptionsField('Template Options', [
-        //     'optionName' => 'template_option',
-        //     'options' => [
-        //         'Provide your own template url.',
-        //         'Choose a template from the catalog.',
-        //         'No template at this time.',
-        //     ],
-        //     'description' => 'Choose a template, provide a url or choose not to use one.',
-        //     'includeAsOption' => false,
-        //     ]);
-        //   $fields['catalog_url'] = new OptionsField('Catalog', [
-        //     'optionName' => 'catalog_url',
-        //     'conditions' => [
-        //         'template_option' => [
-        //             'Choose a template from the catalog.'
-        //         ],
-        //     ],
-        //     'description' => 'The template from which to create your project or your own blank project.',
-        //     'options' => $this->getAvailableCatalog(),
-        //     'asChoice' => FALSE,
-        //     'optionsCallback' => function () {
-        //         return $this->getAvailableCatalog(true);
-        //         },
-        //     ]);
-        //   $fields['template_url'] = new UrlField('Template URL', [
-        //     'optionName' => 'template_url',
-        //     'conditions' => [
-        //         'template_option' => [
-        //             'Provide your own template url.'
-        //         ],
-        //     ],
-        //     'description' => 'The template url',
-        //     'questionLine' => 'What is the URL of the template?',
-        //   ]);  
           $fields['region'] = new OptionsField('Region', [
             'optionName' => 'region',
             'description' => 'The region where the project will be hosted',
@@ -368,18 +333,6 @@ EOF
                 return is_numeric($value) && $value > 0 && $value < 1024;
             },
           ]);
-        //   $fields['initialize'] = new BooleanField('Initialize', [
-        //     'optionName' => 'initialized',
-        //     'conditions' => [
-        //         'template_option' => [
-        //             'Provide your own template url.',
-        //             'Choose a template from the catalog.',
-        //         ],
-        //     ],
-        //     'description' => 'Initialize this environment?',
-        //     'questionLine' => 'Initialize this environment?',
-        //   ]);
-
         return $fields;
     }
 
@@ -430,7 +383,6 @@ EOF
             'questionLine' => 'What is the URL of the template?',
             ]);  
         }
-        
         $fields['initialize'] = new BooleanField('Initialize', [
         'optionName' => 'initialized',
         'conditions' => [
--- a/src/Command/Project/ProjectCreateCommand.php
+++ b/src/Command/Project/ProjectCreateCommand.php
@@ -33,7 +33,8 @@ class ProjectCreateCommand extends CommandBase
         $this->form->configureInputDefinition($this->getDefinition());
 
         $this->addOption('check-timeout', null, InputOption::VALUE_REQUIRED, 'The API timeout while checking the project status', 30)
-            ->addOption('timeout', null, InputOption::VALUE_REQUIRED, 'The total timeout for all API checks (0 to disable the timeout)', 900);
+            ->addOption('timeout', null, InputOption::VALUE_REQUIRED, 'The total timeout for all API checks (0 to disable the timeout)', 900)
+            ->addOption('template', null, InputOption::VALUE_OPTIONAL, 'Choose a starting template or provide a url of one.', false);
 
         $this->setHelp(<<<EOF
 Use this command to create a new project.
@@ -62,6 +63,20 @@ EOF
 
         $options = $this->form->resolveOptions($input, $output, $questionHelper);
 
+        if ($input->getOption('template') !== false) {
+            if (empty(parse_url($input->getOption('template'), PHP_URL_PATH))) {
+                $temp_provided = true;
+            }
+            else {
+                $temp_provided = false;
+            }
+            $this->template_form = Form::fromArray($this->getTemplateFields($temp_provided));
+            $this->template_form->configureInputDefinition($this->getDefinition());
+            $template_options = $this->template_form->resolveOptions($input, $output, $questionHelper);
+    
+        }
+
+
         $estimate = $this->api()
             ->getClient()
             ->getSubscriptionEstimate($options['plan'], $options['storage'], $options['environments'], 1);
@@ -82,11 +97,11 @@ EOF
         }
 
         // Grab the url of the yaml file.
-        if (!empty($options['catalog_url'])) {
-            $options['catalog'] = $options['catalog_url'];
+        if (!empty($template_options['catalog_url'])) {
+            $options['catalog'] = $template_options['catalog_url'];
         }
-        else if (!empty($options['catalog_url'])) {
-            $options['catalog'] = $options['template_url'];
+        else if (!empty($template_options['catalog_url'])) {
+            $options['catalog'] = $template_options['template_url'];
         }
         
         $subscription = $this->api()->getClient()
@@ -157,7 +172,7 @@ EOF
             return 1;
         }
 
-        if ($options['initialize']) {
+        if ($template_options['initialize']) {
             // Use the existing initialize command.
             $project = $this->selectProject($subscription->project_id);
             $environment = $this->api()->getEnvironment('master', $project, null, true);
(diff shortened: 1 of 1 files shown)


src/Command/Project/ProjectCreateCommand.php
```diff
@@ -60,9 +63,24 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $options = $this->form->resolveOptions($input, $output, $questionHelper);
 
+        if ($input->getOption('template') !== false) {
+            if (empty(parse_url($input->getOption('template'), PHP_URL_PATH))) {
+                $temp_provided = true;
+            }
+            else {
+                $temp_provided = false;
+            }
+            $this->template_form = Form::fromArray($this->getTemplateFields($temp_provided));
```
bbujisic: field created dynamically, maybe add it as a `protected $template_form` to the class?

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -78,13 +96,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
+        // Grab the url of the yaml file.
+        if (!empty($template_options['catalog_url'])) {
+            $options['catalog'] = $template_options['catalog_url'];
+        }
+        else if (!empty($template_options['catalog_url'])) {
+            $options['catalog'] = $template_options['template_url'];
+        }
```
bbujisic: the block of above 3 lines (103-105) is a copy of the previous condition (lines 100-102)

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -146,14 +172,30 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
-        $this->stdErr->writeln("The project is now ready!");
+        if ($template_options['initialize']) {
+            // Use the existing initialize command.
+            $project = $this->selectProject($subscription->project_id);
```
pjcdawkins:
```suggestion
            $project = $this->api()->getProject($subscription->project_id);
```

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -146,14 +172,30 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            $environment = $this->api()->getEnvironment('master', $project, null, true);
```
pjcdawkins:
```suggestion
            $environment = $this->api()->getEnvironment('master', $project);
```

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -146,14 +172,30 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            $environment->initialize($subscription->project_options['initialize']['profile'], $subscription->project_options['initialize']['repository']);
```
pjcdawkins: Are we certain at this point that the subscription will contain this data? always? how certain?

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -247,23 +317,85 @@ protected function getFields()
...
+                'options' => [
+                    'Provide your own template url.',
+                    'Choose a template from the catalog.',
+                    'No template at this time.',
+                ],
+                'description' => 'Choose a template, provide a url or choose not to use one.',
+                'includeAsOption' => false,
+            ]);
+    
+            $fields['catalog_url'] = new OptionsField('Catalog', [
```
pjcdawkins: It's not a "catalog URL", it's a template URL... this needs to be a bit clearer, even

```suggestion
            $fields['template_url_from_catalog'] = new OptionsField('Template (from a catalog)', [
```

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -216,29 +258,57 @@ protected function getAvailableRegions($runtime = false)
...
+     * Return the catalog.
+     *
+     * The default list is in the config `service.catalog`. This is
+     * replaced at runtime by an API call.
+     *
+     * @param bool $runtime
+     *
+     * @return array
+     */
+    protected function getAvailableCatalog($runtime = false)
```
pjcdawkins: Maybe `getDefaultCatalog()` ?

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -216,29 +258,57 @@ protected function getAvailableRegions($runtime = false)
...
+    {
+        if ($runtime) {
+            $catalog = [];
+            foreach ($this->api()->getClient()->getCatalog()->getData() as $item) {
```
pjcdawkins: This needs to handle `getCatalog()` returning no items, or returning no items that contain an 'info' and a 'template'.

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -216,29 +258,57 @@ protected function getAvailableRegions($runtime = false)
...
     /**
      * Returns a list of ConsoleForm form fields for this command.
      *
      * @return Field[]
      */
     protected function getFields()
     {
-        return [
-          'title' => new Field('Project title', [
+        $fields = [];
```
pjcdawkins: these format changes to `getFields()` are unnecessary

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -247,23 +317,85 @@ protected function getFields()
...
+        }
+        $fields['initialize'] = new BooleanField('Initialize', [
+        'optionName' => 'initialized',
+        'conditions' => [
+            'template_option' => [
+                'Provide your own template url.',
+                'Choose a template from the catalog.',
+            ],
+        ],
+        'description' => 'Initialize this environment?',
```
pjcdawkins:
```suggestion
        'description' => 'Initialize this project from a template',
```

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -247,23 +317,85 @@ protected function getFields()
...
+        'questionLine' => 'Initialize this environment?',
```
pjcdawkins:
```suggestion
        'questionLine' => 'Initialize this project from the template?',
```

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -247,23 +317,85 @@ protected function getFields()
...
+            ],
+            'description' => 'The template from which to create your project or your own blank project.',
+            'options' => $this->getAvailableCatalog(),
+            'asChoice' => FALSE,
+            'optionsCallback' => function () {
+                return $this->getAvailableCatalog(true);
+                },
+            ]);
+
+            $fields['template_url'] = new UrlField('Template URL', [
```
pjcdawkins: Should we validate this? like download the template, and make sure it contains `profile` and `repository` if we are going to need it for initialization?

src/Command/Project/ProjectCreateCommand.php
shawnawsu: It's not a straight copy paste, one sets catalog_url and one sets template_url.

src/Command/Project/ProjectCreateCommand.php
shawnawsu: update made

src/Command/Project/ProjectCreateCommand.php
shawnawsu: update made

src/Command/Project/ProjectCreateCommand.php
shawnawsu: update made

src/Command/Project/ProjectCreateCommand.php
shawnawsu: reverted

src/Command/Project/ProjectCreateCommand.php
shawnawsu: update made

src/Command/Project/ProjectCreateCommand.php
shawnawsu: update made

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -60,9 +63,24 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
+        if ($input->getOption('template') !== false) {
+            if (empty(parse_url($input->getOption('template'), PHP_URL_PATH))) {
+                $temp_provided = true;
+            }
+            else {
+                $temp_provided = false;
+            }
+            $this->template_form = Form::fromArray($this->getTemplateFields($temp_provided));
+            $this->template_form->configureInputDefinition($this->getDefinition());
```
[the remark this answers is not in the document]
shawnawsu: removed

src/Command/Project/ProjectCreateCommand.php
shawnawsu: Makes sense

src/Command/Project/ProjectCreateCommand.php
shawnawsu: I've plopped in some checks to make sure required information is coming through. I'm not wedded to the error message though, so please do review.

src/Command/Project/ProjectCreateCommand.php
shawnawsu: good idea.

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -57,12 +63,24 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $template = $input->getOption('template');
+        
+        if ($template !== false) {
+            if (empty(parse_url($template, PHP_URL_PATH))) {
+                $temp_provided = true;
+            }
+            else {
+                $temp_provided = false;
+            }
+            $this->template_form = Form::fromArray($this->getTemplateFields($temp_provided));
```
pjcdawkins: No need for this to be a class property, it can be a regular variable

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -78,13 +96,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
+        // Grab the url of the yaml file.
+        if (!empty($template_options['catalog_url'])) {
```
pjcdawkins: this now needs to be `template_url_from_catalog`, I think

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -78,13 +96,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            $options['catalog'] = $template_options['catalog_url'];
+        }
+        else if (!empty($template_options['catalog_url'])) {
```
pjcdawkins:
```suggestion
        else if (!empty($template_options['template_url'])) {
```

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -78,13 +96,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
+        // Grab the url of the yaml file.
+        if (!empty($template_options['catalog_url'])) {
+            $options['catalog'] = $template_options['catalog_url'];
```
pjcdawkins:
```suggestion
            $options['catalog'] = $template_options['template_url_from_catalog'];
```

config.yaml
```diff
@@ -80,6 +80,8 @@ service:
     - standard
     - medium
     - large
+  catalog:
+    - https://github.com/platformsh/template-drupal7-vanilla/blob/master/.platform.template.yaml
```
Crell: This file has been removed.  (Or, well, it's supposed to have been removed. We may have some cleanup to do yet.)  The template.yaml files are no longer included in the Git repository since they don't have any real value there.  They're in the template-builder repository for the first-party templates.

Also, a catalog is a collection of such template definitions, no?  Not a single entry?

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -216,6 +262,40 @@ protected function getAvailableRegions($runtime = false)
...
+            $catalog = [];
+            $catalog_items = $this->api()->getClient()->getCatalog()->getData();
+            if (!empty($catalog_items)) {
+                foreach ($catalog_items as $item) {
+                    if (isset($item['info']) && isset($item['template'])) {
+                        $catalog[$item['template']] = $item['info']['name'];
+                    }
+                }
+            }
+            $catalog['empty'] = 'Empty Project';
```
pjcdawkins: It seems you want this always to be an option in `$catalog` whether it comes from config or from the remote catalog

Crell:
If I'm reading this properly, the `--template` flag serves double duty: Either provide a template (in which case it's a short-circuit way to also run `initialize`) or provide a catalog to pick from that will interactively ask which project you want.  Am I reading it correctly?

If so, then IMO it should be split.  Have a `--template` that specifies a Git URL, which then short-circuits `initialize`, and have a separate `--catalog` which specifies a catalog definition that will interactively ask which template you want.  But those behaviors should not be folded into a single switch.  They're separate activities.

config.yaml
Kazanir: I believe we should not need a default catalog file here. This should be using the `/setup/catalog` API and the API is responsible (based on the user's OEM settings, vendor, and organization) for providing the correct default catalog file.

Kazanir: Looks like this is provided solely as a fallback in the case that the CLI has no runtime in which case it should be `http://accounts.platform.sh/sites/default/files/catalog/catalog-psh.yaml`.

src/Command/Project/ProjectCreateCommand.php
Kazanir: I don't think so. We're validating the template URL in the API call now, and should be catching failed API calls anyway. If it lacks the initialize key (which is perfectly fine to do) then we should just skip initialization.

Kazanir: I guess the thing to do here is make sure we're validating the form entry by submitting the newly-chosen template URL to `/setup/options` API?

Kazanir:
> If I'm reading this properly, the `--template` flag serves double duty: Either provide a template (in which case it's a short-circuit way to also run `initialize`) or provide a catalog to pick from that will interactively ask which project you want. Am I reading it correctly?
> 
> If so, then IMO it should be split. Have a `--template` that specifies a Git URL, which then short-circuits `initialize`, and have a separate `--catalog` which specifies a catalog definition that will interactively ask which template you want. But those behaviors should not be folded into a single switch. They're separate activities.

We're skipping the "catalog definition" step, that's provided by Accounts based on any of your vendor/organization[NYI]/user-level catalog files, or falling back to the default if none of those are found. The `--template` key here (if I understand the MR right) will take you into the catalog process if you supply it without an argument for a specific template file, but the catalog itself is being provided by who you are via Accounts, not the CLI flag.

Console is going to support a `?catalog=` parameter for the sake of "Try Our Stuff On Platform" buttons but there's really no equivalent CLI functionality. Scripting the `?template` input to a specific template file is a pretty likely use case though.

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -146,14 +169,37 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
-        $this->stdErr->writeln("The project is now ready!");
+        if ($template_options['initialize']) {
```
Kazanir: Doesn't this need to be the $subscription->project_options at this point in the process?

shawnawsu: Not unless we're automatically initializing projects. This is looking for the `initialize` flag then from line 176 it's testing to make sure all the required parameters are being passed to successfully initialize the project.

src/Command/Project/ProjectCreateCommand.php
shawnawsu: I'm really not sure about this, so I removed it. Now that we're falling back on to the `setup/options` api this is probably a non-issue.

config.yaml
shawnawsu: That was just a place holder for testing, I've updated it to point to the correct file.

src/Command/Project/ProjectCreateCommand.php
shawnawsu: The yaml now gets error checked using the `setup/options` api.

src/Command/Project/ProjectCreateCommand.php
shawnawsu: I've updated the process to automatically initialize any project that contains the required info to successfully initialize it.

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -216,6 +273,39 @@ protected function getAvailableRegions($runtime = false)
...
+                        $catalog[] = $item['template'];
+                    }
+                }
+            }
+        } else {
+            $catalog = (array) $this->config()->get('service.catalog');
+        }
+
+        if (empty($catalog)) {
+            // Should we throw an error here? Do we want to kill the process?
```
bbujisic: What happens if the `getTemplateFields()` receives an empty catalog array? If nothing significant -- I'd remove this condition. If something unexpected -- I'd throw an Exception and bail out.

I don't have a clue what am I writing about, right here, so please be free to ignore me if I make no sense at all.

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -203,11 +250,21 @@ protected function getAvailablePlans($runtime = false)
     protected function getAvailableRegions($runtime = false)
     {
         if ($runtime) {
-            $regions = [];
-            foreach ($this->api()->getClient()->getRegions() as $region) {
-                if ($region->available) {
-                    $regions[$region->id] = $region->label;
+            // First look to the setup/options API for a region list.
+            $account = $this->api()->getMyAccount(true);
+            $setupOptions = $this->api()->getClient()->getSetupOptions(NULL, NULL, NULL, $account['username'], NULL);
```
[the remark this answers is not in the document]
shawnawsu: Yes, you're absolutely right but right now the `setup/options` api isn't in production so these fall-backs allow you to at least test the passing of the template and initialize parameters.

src/Command/Project/ProjectCreateCommand.php
shawnawsu: If `$catalog` comes back as empty it will cause a logic exception since it's an `asChoice` field. So, do we want it to fail and skip to the next step and just now allow a template choice or stop the process? I'm not sure what the preference is.

bbujisic: Hm, I'd personally fail it. What do you think @pjcdawkins?

pjcdawkins: I'd return `[]` here, and let Symfony throw the LogicException

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -216,6 +273,39 @@ protected function getAvailableRegions($runtime = false)
...
+     * @param bool $runtime
+     *
+     * @return array
+     */
+    protected function getDefaultCatalog($runtime = false)
+    {
+        if ($runtime) {
+            $catalog = [];
+            $catalog_items = $this->api()->getClient()->getCatalog()->getData();
+            if (!empty($catalog_items)) {
```
pjcdawkins: the `if` isn't needed - if it's always an array, the `foreach` won't mind if it's empty

PR #832, opened 2019-08-07
Add --charset option to the db:dump command

pjcdawkins:
```
# download this build
curl https://pr-832-kgcph7y-mpaw4nkgdwaea.eu-2.platformsh.site/platform.phar -o ./platform.phar

# create a dump with the utf8mb4 character set encoding
php platform.phar db:dump --charset utf8mb4
```


bircher:
+1
that works for our use case.

PR #846, opened 2019-08-29
bugfixing some PE customers

matthiaz:
some PE clusters do not have the  PROCESS privilege(s) and thus, have no access to the sys_tablespaces, revert to legacy way

### DIFF
--- a/src/Command/Db/DbSizeCommand.php
+++ b/src/Command/Db/DbSizeCommand.php
@@ -363,7 +363,7 @@ class DbSizeCommand extends CommandBase
             $this->debug('Checking InnoDB separately for more accurate results...');
             try {
                 $innoDbSize = $host->runCommand($this->mysqlInnodbQuery($database));
-            }catch(Throwable $e) {//some PE clusters do not have the  PROCESS privilege(s) and thus, have no access to the sys_tablespaces, revert to legacy way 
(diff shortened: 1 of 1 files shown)


src/Command/Db/DbSizeCommand.php
```diff
@@ -361,7 +361,11 @@ private function getMySqlUsage(HostInterface $host, array $database) {
         $innoDbSize = 0;
         if ($allocatedSizeSupported) {
             $this->debug('Checking InnoDB separately for more accurate results...');
-            $innoDbSize = $host->runCommand($this->mysqlInnodbQuery($database));
+            try {
+                $innoDbSize = $host->runCommand($this->mysqlInnodbQuery($database));
+            }catch(Throwable $e) {//some PE clusters do not have the  PROCESS privilege(s) and thus, have no access to the sys_tablespaces, revert to legacy way
```
pjcdawkins: This catch should be more precise - the exception will be a `\Symfony\Component\Process\Exception\RuntimeException`, and the message will mention error code `1227`

pjcdawkins: (and mainly, `Throwable` is a PHP 7 only thing, the CLI supports PHP 5.5.9+)

matthiaz: You are absolutely right. Changed and pushed again.

PR #847, opened 2019-09-02
Stop showing a redeploy warning when embedded activities aren't provided

pjcdawkins:
Activities are no longer embedded in the response (the Result) for actions that
trigger activities. We therefore cannot use the absence of activities as
evidence that the environment has not been automatically redeployed.


pjcdawkins:
This might be considered a bug in the API

pjcdawkins:
An API fix has been implemented

Issue #865, opened 2019-10-29
SSH command enterprise problems


cs278:
I've just updated my version of the CLI and encountered a number of issues using the SSH command which seem to be related to #800. Unfortunately these are deal breakers for me so I've had to revert to an older version.

```
$ platform ssh -e production
Enter a number to choose the app or a worker:
  [0] myapp (default)
  [1] myapp--api-worker-1
  [2] myapp--api-worker-2
> 
```

If I hit enter at this point I get `Value "" is invalid`, the default indicator seems to be indicating nothing.

If I select 1 or 2 (which are workers) I get:

```
  [RuntimeException]                                                                   
  Unable to find the SSH URL for the app "myapp" containing the worker "api-worker-1"  
```

This is due to production being an enterprise cluster in my use case.

It does not appear to be possible to bypass this selection using `--app`:

```
$ platform ssh -e production --app myapp 
Enter a number to choose the app or a worker:
  [0] myapp (default)
  [1] myapp--api-worker-1
  [2] myapp--api-worker-2
 > 
```

pjcdawkins:
You can add --yes as a workaround - the selection will be bypassed

cs278:
I think everything is addressed.

PR #866, opened 2019-10-30
Allow --app option to correctly select application

cs278:
This fixes the last problem identified in #865, `platform ssh --app myapp` does not actually select the application to connect to.


pjcdawkins:
Works and looks good to me, thanks @cs278 !

PR #867, opened 2019-10-30
Fix question helper

cs278:
Fixes issue identified in #865 where default selection is not respected.


pjcdawkins:
Looks good to me, thanks @cs278 !

PR #868, opened 2019-10-30
Hide worker containers on enterprise

cs278:
Fixes last remaining problem raised in #865 by disabling worker selection in enterprise environments.

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -886,11 +886,6 @@ abstract class CommandBase extends Command implements MultiAwareInterface
 
         $environment = $this->getSelectedEnvironment();
 
-        // Enterprise environments do not have separate containers for workers.
-        if ($includeWorkers && $environment->deployment_target === 'enterprise') {
-            $includeWorkers = false;
-        }
-
         try {
             $deployment = $this->api()->getCurrentDeployment(
                 $environment,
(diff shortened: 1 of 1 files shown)


src/Command/CommandBase.php
```diff
@@ -885,6 +885,12 @@ protected function selectRemoteContainer(InputInterface $input, $includeWorkers
         }
 
         $environment = $this->getSelectedEnvironment();
+
+        // Enterprise environments do not have seperate containers for workers.
+        if ($includeWorkers && $environment->deployment_target === 'enterprise') {
```
pjcdawkins: Unfortunately the deployment_target for an Enterprise (dedicated) environment isn't necessarily the string `enterprise`. Elsewhere we check for `!== 'local'`, which is hacky but works.

src/Command/CommandBase.php
```diff
@@ -885,6 +885,12 @@ protected function selectRemoteContainer(InputInterface $input, $includeWorkers
         }
 
         $environment = $this->getSelectedEnvironment();
+
+        // Enterprise environments do not have seperate containers for workers.
```
pjcdawkins:
```suggestion
        // Enterprise environments do not have separate containers for workers.
```

src/Command/CommandBase.php
pjcdawkins: For this reason - the slightly unreliable hack - I want to hide the interactive worker choice for non-`local` environments, but still respect the `--worker` option.

cs278: @pjcdawkins updated, I haven't tested it however.

PR #906, opened 2020-02-14
Pre commit hooks

matthiaz:


matthiaz:
@pjcdawkins do you need further input from me on this?
I'm looking back into this and it still has its uses.

PR #913, opened 2020-04-07
Remove --from-address for health.email integrations, as there is now a good API default

pjcdawkins:


pjcdawkins:
Closing in favour of #915

PR #915, opened 2020-04-07
Remove default From address for health.email notifications

pjcdawkins:
The API now provides a sensible default

### DIFF
--- a/src/Command/Integration/IntegrationCommandBase.php
+++ b/src/Command/Integration/IntegrationCommandBase.php
@@ -342,7 +342,7 @@ abstract class IntegrationCommandBase extends CommandBase
                 'conditions' => ['type' => [
                     'health.email',
                 ]],
-                'description' => 'The From address for alert emails',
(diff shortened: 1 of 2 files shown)


MariadeAnton:
Tested this and it works like a charm, thanks for updating the CLI so quickly @pjcdawkins

src/Command/Integration/IntegrationCommandBase.php
```diff
@@ -344,6 +344,7 @@ private function getFields()
                 ]],
                 'description' => 'The From address for alert emails',
```
MariadeAnton, line 345:
```suggestion
                'description' => '[Optional] Custom From address for alert emails',
```
What do you think about something along these lines @pjcdawkins ? I'd like the description text to help folks understand they can add a custom one, but that it's not required :)

pjcdawkins: Yes good point

PR #917, opened 2020-04-09
Fix deleting httpaccess settings using a merge patch

pjcdawkins:


src/Command/Environment/EnvironmentHttpAccessCommand.php
```diff
@@ -174,7 +174,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
             // Merge existing settings. Not using a reference here, as that
             // would affect the comparison with $current later.
             foreach ($current as $key => $value) {
-                if (!isset($accessOpts[$key])) {
+                // Use null to delete an existing value.
+                if ($accessOpts[$key] === [] && is_array($value) && $value !== []) {
```
ralt, line 178: Aren't you going to potentially get a warning  if you don't do the if !isset() first?

pjcdawkins: true

PR #823, opened 2019-07-12
WIP: Add script integration

pjcdawkins:
As

```bash
platform integration:add --type script --file path/to/script.js
```

Uses https://github.com/platformsh/console-form/blob/master/src/Field/FileField.php

PR #919, opened 2020-04-15
Add script integration

pjcdawkins:
Successor to #823


Crell:
The CLI help text currently says that the default events to listen to are `["*"`].  Incorrect.  The default is `environment.push`.  The "all" value is just plain `*`, it seems.

Crell:
Request: In the `integration:activity:list` table, include the event name that triggered that run. (That makes it easier to tell which one it is, otherwise all of the runs look identical.)

Crell:
Request: A command to show "most recent activity's log".  Otherwise every time you trigger a new activity to test some change you have to look up its ID manually.  Maybe just a `--latest` flag on commands that take an activity ID?

Crell:
The description field is rather useless in the table output.  It's always just "ran this type of integration", which since you're showing only one integration will always be the same string over and over again.  It can be removed to save space.

Crell:
`DEPRECATED The option --type is deprecated and will be removed in a future version`

I keep getting this message when I update a script, but without the `--type` property it fails.  Let's pick one. :smile:

pjcdawkins:
> The CLI help text currently says that the default events to listen to are `["*"]`. Incorrect. The default is `environment.push`. The "all" value is just plain `*`, it seems.

The CLI wasn't sending the default. I've fixed this.

> Request: In the `integration:activity:list` table, include the event name that triggered that run. (That makes it easier to tell which one it is, otherwise all of the runs look identical.)

I could explore that in another issue.

> Request: A command to show "most recent activity's log". Otherwise every time you trigger a new activity to test some change you have to look up its ID manually. Maybe just a `--latest` flag on commands that take an activity ID?

The `activity:get` and `activity:log` commands (and integration equivalents) already find the latest activity by default. BTW I'd love to change this to the "latest pending activity" by default, but the API doesn't allow for filtering activities by state yet. Anyway, this is not related to the current PR.

> The description field is rather useless in the table output. It's always just "ran this type of integration", which since you're showing only one integration will always be the same string over and over again. It can be removed to save space.

Integrations are free to populate the description field with more detail, that's a concern for the API (or at least, not this PR).

> `DEPRECATED The option --type is deprecated and will be removed in a future version`
> 
> I keep getting this message when I update a script, but without the `--type` property it fails. Let's pick one.

Fixed, I think.

Crell:
I still get the DEPRECATED message if I use `--type` on an `integration:update` command.

pjcdawkins:
> I still get the DEPRECATED message if I use --type on an integration:update command.

That's intentional

Crell:
Then how are you supposed to set the type for an integration if not with the switch?

pjcdawkins:
> Then how are you supposed to set the type for an integration if not with the switch?

If you're running `integration:update`, the integration already exists, you can't change the type.

Crell:
Ahhh... fair.  The error message is not descriptive for that, though.  (If it confused me, it will confuse lots of customers.)  Suggested alternative: "The --type switch is not supported on updates.  You must remove and re-add an integration to change it to a different type."

PR #921, opened 2020-04-16
replace foo/bar example with username/repository

MariadeAnton:
If username/repository doesn't work, we could go with a simplified version:

- user/repo
- username/repo

what do you think? Thanks!


MariadeAnton:
Great catch, @pjcdawkins, thank you  - just updated it to `owner/repository` which kinda covers the 3 cases better, wdyt?

PR #922, opened 2020-04-20
Add an alias for environment deactivation.

Crell:
"Deactivate" is the mirror of "Activate," and is the terminology used in the UI anyway.


pjcdawkins:
We should add a warning similar to Console's before doing this.

Screenshot from 2020-04-20 15-12-23

Crell:
It's just an alias, so it would have all the same warnings that "delete" has now, yes?

PR #932, opened 2020-05-26
Support grouped applications.yaml configuration

pjcdawkins:


xtfer:
Since I dont have a local copy of the CLI, I tested this with a zip off the source branch, and was able to run `platform build` successfully. Is there anything else you want tested?

pjcdawkins:
That's great thanks @xtfer

PR #937, opened 2020-05-29
Change backup:list (backups) command to new API

pjcdawkins:
Makes the `restorable` property available


pjcdawkins:
The API is insufficient as it sorts backups by oldest first

PR #939, opened 2020-06-02
Refactor API token configuration to remove circular dependency

pjcdawkins:


pjcdawkins:
- [ ] split into separate service

PR #941, opened 2020-06-03
Automatically select an SSH key corresponding to the user account

pjcdawkins:


pjcdawkins:
- [ ] write into SSH config (split this from Certifier service)
- [ ] one session's SSH config file at a time

Issue #955, opened 2020-07-13
No output without iconv extension


spk:
Hi there,

Without this extension running the `platform workers` command gives
you an output like:

```
Workers on the project xxx (xxx), environment Master (master):
```

But without any output of the workers or error. Running it with verbose
mode give you more info:

```
    Fatal error: Uncaught Error: Call to undefined function Symfony\Polyfill\Mbstring\iconv() in phar:///root/.platformsh/bin/platform/vendor/symfony/polyfill-mbstring/Mbstring.php:713
    Stack trace:
     #0 phar:///root/.platformsh/bin/platform/vendor/symfony/polyfill-mbstring/bootstrap.php(102): Symfony\Polyfill\Mbstring\Mbstring::mb_strwidth('app--queue', 'ASCII')
     #1 phar:///root/.platformsh/bin/platform/vendor/symfony/console/Helper/Helper.php(54): mb_strwidth('app--queue', 'ASCII')
     #2 phar:///root/.platformsh/bin/platform/vendor/symfony/console/Helper/Helper.php(123): Symfony\Component\Console\Helper\Helper::strlen('app--queue')
     #3 phar:///root/.platformsh/bin/platform/src/Console/AdaptiveTable.php(353): Symfony\Component\Console\Helper\Helper::strlenWithoutDecoration(Object(Symfony\Component\Console\Formatter\OutputFormatter), 'app--queue')
     #4 phar:///root/.platformsh/bin/platform/src/Console/AdaptiveTable.php(267): Platformsh\Cli\Console\AdaptiveTable->getCellWidth('app--queue')
     #5 phar:///root/.platformsh/bin/platform/src/Co in phar:///root/.platformsh/bin/platform/vendor/symfony/polyfill-mbstring/Mbstring.php on line 713
```

env: LANG=C.UTF-8
Adding the `php-iconv` extension fixed the issue and I could see the workers list

I've noticed that this extension was checked before on the installer but then removed

Cheers,
Laurent

pjcdawkins:
It looks like the `mbstring` extension would solve this too. The installer should be made to check for one of them.

PR #958, opened 2020-07-20
README: add extensions needed by cli

spk:
ref https://github.com/platformsh/platformsh-cli/issues/955

This add more info on the readme about required extensions
Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@ The **Platform.sh CLI** is the official command-line interface for [Platform.sh]
 ## Requirements
 
 * Operating system: Linux, OS X, or Windows 10
-* PHP 5.5.9 or higher, with cURL support
+* PHP 5.5.9 or higher, with the following extensions: `curl`, `json`,
+ `mbstring`, `pcre`, and `phar`. The installation command will check for these.
 * Git
 * A Bash-like shell:
   * On OS X or Linux/Unix: SH, Bash, Dash or ZSH - usually the built-in shell will work.


README.md
```diff
@@ -5,7 +5,7 @@ The **Platform.sh CLI** is the official command-line interface for [Platform.sh]
 ## Requirements
 
 * Operating system: Linux, OS X, or Windows 10
-* PHP 5.5.9 or higher, with cURL support
+* PHP 5.5.9 or higher, with `cURL` `json` `phar` `mbstring` support
```
pjcdawkins, line 8:
```suggestion
* PHP 5.5.9 or higher, with the following extensions: `curl`, `json`, `mbstring`, `pcre`, and `phar`. The installation command will check for these.
```

with some notes

- `phar` is only required for the recommended installation method, it's not technically a hard requirement (e.g. you could install the CLI via Composer).
- I believe `curl` support is not really a hard requirement, the CLI uses Guzzle more or less throughout which can fall back to the built-in `fopen` transport. But I haven't tried the CLI without curl. I think it makes certain things more difficult (like configuring TLS root certs?).
- one of `mbstring` or `iconv` is required, apparently for Symfony Console. I think `mbstring` is newer and more widely available/preinstalled.
- thankfully `json` will always be included in PHP 8.

spk:
Strange I cannot reply to the thread... Thanks for the suggestion and the info I've amended the commit

pjcdawkins:
Thanks @spk!

PR #971, opened 2020-08-12
Disable use of SSH agent unless explicitly enabled in the user config file

pjcdawkins:
Add this to the user config file to restore the old behavior (automatically add keys/certificates to the SSH agent):

```yaml
# ~/.platformsh/config.yaml
api:
  add_to_ssh_agent: true
```

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -145,6 +145,10 @@ api:
   # Can be replaced by a single value using the {application.env_prefix}SSH_DOMAIN_WILDCARD env var.
   ssh_domain_wildcards: ['*.platform.sh']
 
+  # Whether auto-generated SSH certificate identities should be added to the SSH agent.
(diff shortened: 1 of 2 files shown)


pjcdawkins:
Reasoning:

* Use of the SSH agent is not necessary to use SSH certificates, except in the case of agent forwarding (`-A`), which is rare, and should be discouraged.
* Those who know what they're doing well enough to use `-A`, know how to edit `~/.platformsh/config.yaml`
* The SSH agent is "dumb" and sends platform.sh-specific SSH certificates to irrelevant SSH hosts (it can't be controlled by a `Host` or `Match` directive).
* Apparently the GnuPG SSH agent does not properly remove expired or deleted keys. Regeneration every hour causes an ever-growing list of keys.

thoughts, @ralt ?

PR #980, opened 2020-08-26
Generate a cleaner ssh config file

SphericalHam:
These changes make the config file generated a bit cleaner in some cases. It's also now possible to determine that no config need be generated at all.

### DIFF
--- a/src/Service/SshConfig.php
+++ b/src/Service/SshConfig.php
@@ -85,15 +85,6 @@ class SshConfig {
             $lines[] = '';
         }
 
-        $sessionSpecificFilename = $this->getSessionSshDir() . DIRECTORY_SEPARATOR . 'config';
-        $includerFilename = $this->getCliSshDir() . DIRECTORY_SEPARATOR . 'session.config';
-        if (empty($lines)) {
-            if (\file_exists($includerFilename) || \file_exists($sessionSpecificFilename)) {
-                $this->fs->remove([$includerFilename, $sessionSpecificFilename]);
-            }
-            return false;
-        }
-
         // Add default files if there is no preferred session identity file.
         if ($sessionIdentityFile === null && ($defaultFiles = $this->getUserDefaultSshIdentityFiles())) {
             $lines[] = '# Include SSH "default" identity files:';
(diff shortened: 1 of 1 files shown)


pjcdawkins:
Theoretically none of the `Host` lines should be necessary, because the SSH manual says:

https://www.freebsd.org/cgi/man.cgi?ssh_config(5)

> **Include** directive may appear inside a **Match** or **Host**
>             block to perform conditional inclusion.

and the root `~/.ssh/config` already wraps the Include inside a Host block.

I'm not sure if this conditional inclusion actually works.

SphericalHam:
You're quite right, the entire file is only included for matching hosts - a quick look at the openssh source confirms that it does indeed work that way. 

Personally, I think it's wise to avoid assuming that root ssh config is set in a particular way - the user is asked if the config should be modified, after all.

One thing I do with my ssh config files is to avoid inclusions at all. I do this because paramiko, the python ssh library, doesn't support inclusions in config files at all. Generally, when I manage config files, I use a script to join all config fragments together to form the final config file. This approach can play poorly with config fragments that have an empty host block at the end.

If we do assume that the file is included conditionally, we could do away with generating host blocks altogether - just use the `match exec`. I can re-work this PR for that if that's what we prefer.

pjcdawkins:
The match exec is only there if there is a cert.

Sounds like you'd need the file to end in `Host *`, then there isn't a risk from an empty host block? As well as the tidying you suggest

SphericalHam:
I guess my thinking is kinda backwards.

If the file is included with openssh's `Include` directive, then it doesn't matter what Host blocks there are in the included file - the Host blocks only apply to that file.

If the file is concatenated the way I describe, then it sort of doesn't matter how the file ends - empty host block or not, the next file would have to start with a Host line to set the state to something known.

So I guess, then, that the only thing this does is generate a more tidy file.

pjcdawkins:
The `Host *.platform.sh` within this file (whether it's "empty" or not) will continue applying to any statements except those after another Host or Match, so if you're concatenating any SSH config files, you'll have unpredictable results. That's why it'd need `Host *` or `Match all` at the end

PR #991, opened 2020-10-19
Increase timeout when fetching activity logs

a1phyr:
This is necessary for Testbed which can be slower.

cc @Rivee


pjcdawkins:
This is clearly something that needs fixing upstream, but relatively harmless here

PR #995, opened 2020-10-21
Stop using SSH home directory substitution

pjcdawkins:
This PR is based on a bug found running the CLI inside GitHub Actions:

1. The Platform.sh CLI passes certificate/key filenames to ssh using `%d` to substitute for the home directory, mostly for neatness, and to bypass some shell path quoting issues on Windows (e.g. when the user name contains a space) (https://github.com/platformsh/platformsh-cli/blob/fbb54779c624be881a35ce82da53df07141513ce/src/Service/SshConfig.php#L136-L138)
2. GitHub Actions sets the home directory via the `$HOME` environment variable to `/github/home`
3. The Platform.sh CLI reads `$HOME` as you'd expect
4. However, ssh (OpenSSH) picks up the home directory using's C's `getpwuid(getuid())`, which uses the user database (`/etc/passwd`) directly, and it ignores the `$HOME` environment variable: https://github.com/openssh/openssh-portable/blob/eab2888cfc6cc4e2ef24bd017da9835a0f365f3f/ssh.c#L682
4. Therefore the `ssh` process in GitHub Actions believes the home directory is `/root`, so it can't find the SSH certificate stored in `/github/home`

Reported via chat - https://platformsh.slack.com/archives/C0JHEUHQD/p1603307475289200

### DIFF
--- a/src/Service/SshConfig.php
+++ b/src/Service/SshConfig.php
@@ -133,9 +133,6 @@ class SshConfig {
      */
     public function formatFilePath($path)
     {
-        // Replace the home directory with the SSH token %d, which solves most quoting problems.
-        $home = $this->config->getHomeDirectory();
-        $path = \str_replace($home, '%d', $path);
         if (\strpos($path, ' ') === false) {
             return $path;
         }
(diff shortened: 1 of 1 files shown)


markdorison (approved):
I tested this change by installing this [version of the CLI](https://pr-995-ertlwua-mpaw4nkgdwaea.eu-2.platformsh.site/) in a GitHub action docker container and can confirm it now connects successfully to the specified platform environment.

pjcdawkins:
Thanks. I've tested on Windows 10, with a space in the home directory path, to ensure the quoting still works as intended.

PR #1008, opened 2020-12-08
Make SSH connection error more generic

tylerssn:
When booted from an SSH session during deployment users see "You probably need to add an SSH key, with: platform ssh-key:add".

This proposed change covers disconnects due to deployments.

### DIFF
--- a/src/Service/SshDiagnostics.php
+++ b/src/Service/SshDiagnostics.php
@@ -205,7 +205,7 @@ class SshDiagnostics
                 $this->stdErr->writeln('');
             }
             $this->stdErr->writeln(sprintf(
-                'You probably need to add an SSH key, with: <comment>%s ssh-key:add</comment>',
+                '<error>[Could not connect]</error> Either the app is being deployed or you need to add an SSH key, with: <comment>%s ssh-key:add</comment>',
                 $executable
             ));
             return;


pjcdawkins:
Here's a more convoluted solution: #1009

src/Service/SshDiagnostics.php
```diff
@@ -205,7 +205,7 @@ public function diagnoseFailure($uri, $exitCode, Process $failedProcess = null,
                 $this->stdErr->writeln('');
             }
             $this->stdErr->writeln(sprintf(
-                'You probably need to add an SSH key, with: <comment>%s ssh-key:add</comment>',
+                '<error>[Could not connect]</error> Either the app is being deployed or you need to add an SSH key, with: <comment>%s ssh-key:add</comment>',
```
pjcdawkins, line 208: I'm wondering if there might be more possibilities at this stage. Maybe if we can't provide accurate advice here we should remove the SSH key hint entirely.

pjcdawkins:
This (committed to 3.x) has two measures that should solve the misleading error: https://github.com/platformsh/platformsh-cli/commit/7fa4b45ed27f080bb0dd9c2e4b1858bf3bf1a13b

- the "test connection" is not made if the original connection was too long ago (>3s), to avoid unnecessary slowness and to avoid situations where you can connect fine, but some error happened slightly later due to redeployment or another process exiting with a 255 255 exit code
- it only prints the SSH key message if "Permission denied (publickey)" was in the SSH error output

PR #1033, opened 2021-02-10
installer: support using a GitHub token to download assets

pjcdawkins:


pjcdawkins:
This does not work with the current naive `file_get_contents()`; at least some way of "manually" following redirects and altering the headers on different requests would be necessary, which may be more code than it's worth. GitHub accepts the Authorization header on the initial request, but then complains after the redirect:

```
Only one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Authorization header should be specified
```

PR #1034, opened 2021-02-10
Use http scheme for SOLR service tunnel URL

SerheyDolgushev:


pjcdawkins:
This isn't solely exposed as the tunnel URL, it's also there directly on the project container (and externally) as:

```sh
platform relationships --property solr.0.url
```

if anyone's relying on that starting with `solr://` ([even though they shouldn't](https://xkcd.com/1172/)) then it could break something. Hmm.

SerheyDolgushev:
Sorry, so what would be the best solution here? Maybe there is a way to notify about this change in advance? Or maybe it worth introducing a new method for decorating just tunnel URLs?

PR #1077, opened 2021-08-26
Removing 'master' in platform get command

matthiaz:
Changing 'master' to project->default_branch so we don't mention master if they specified something else.

Also, attempt to move to the correct branch after the init.

### DIFF
--- a/src/Command/Project/ProjectGetCommand.php
+++ b/src/Command/Project/ProjectGetCommand.php
@@ -132,7 +132,7 @@ class ProjectGetCommand extends CommandBase
             }
 
             $this->debug('Initializing the repository');
-            $git->init($projectRoot, true);
+            $git->init($projectRoot, $project->default_branch, true);
 
             $this->debug('Initializing the project');
             $localProject->mapDirectory($projectRoot, $project);
--- a/src/Command/Project/ProjectGetCommand.php
+++ b/src/Command/Project/ProjectGetCommand.php
@@ -137,6 +137,11 @@ class ProjectGetCommand extends CommandBase
             $this->debug('Initializing the project');
             $localProject->mapDirectory($projectRoot, $project);
 
+            if($git->getCurrentBranch($projectRoot) != $project->default_branch) {
+                $this->debug('current branch does not match the default_branch, create it.');
+                $git->checkOutNew($project->default_branch, null, null, $projectRoot);
+            }
+
             $this->stdErr->writeln('');
             $this->stdErr->writeln(sprintf(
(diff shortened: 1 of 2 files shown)


src/Command/Project/ProjectGetCommand.php
```diff
@@ -137,15 +137,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $this->debug('Initializing the project');
             $localProject->mapDirectory($projectRoot, $project);
 
+            if($git->getCurrentBranch($projectRoot) != $project->default_branch) {
+                $this->debug('current branch does not match the default_branch, create it.');
+                $git->checkOutNew($project->default_branch, null, null, $projectRoot);
+            }
```
pjcdawkins, line 143: We should probably try to pass `--initial-branch` to the `git init` call above, maybe only if Git is version 2.28 or above.

https://git-scm.com/docs/git-init/2.28.0

matthiaz: https://github.com/platformsh/platformsh-cli/pull/1077/files#diff-758a78d9bf121fd02f690e51541be9c098d2ad2ee91f1f991e885162b30a2b68R161 like this then?

src/Service/Git.php
```diff
@@ -151,13 +151,17 @@ public function execute(array $args, $dir = null, $mustRun = false, $quiet = tru
      *
      * @return bool
      */
-    public function init($dir, $mustRun = false)
+    public function init($dir, $initial_branch='master', $mustRun = false)
```
pjcdawkins, line 154:
```suggestion
    public function init($dir, $initial_branch = '', $mustRun = false)
```

src/Service/Git.php
```diff
@@ -151,13 +151,17 @@ public function execute(array $args, $dir = null, $mustRun = false, $quiet = tru
...
     {
         if (is_dir($dir . '/.git')) {
             throw new \InvalidArgumentException("Already a repository: $dir");
         }
 
-        return (bool) $this->execute(['init'], $dir, $mustRun, false);
+        $args = ['init'];
+        if ($this->supportsGitInitialBranchFlag()) {
```
pjcdawkins, line 161:
```suggestion
        if ($initial_branch !== '' && $this->supportsGitInitialBranchFlag()) {
```

src/Command/Project/ProjectGetCommand.php
pjcdawkins: Exactly

pjcdawkins:
thanks @matthiaz

PR #1108, opened 2022-01-26
Add country support to org:create and org:info commands

pjcdawkins:


bojanz:
Let's party like it's [2014](https://github.com/platformsh/platformsh-cli/commit/8891716d06e9be591f800c57ed5cb01877cebd86)!

PR #1148, opened 2022-08-11
Add support for pg_dump -n and -N flags

digital2real:
--pg-namespace=PG-NAMESPACE
	translate to -n, --schema=SCHEMA
	dump the named schema(s) only

--pg-exclude-namespace=PG-EXCLUDE-NAMESPACE
	translate to -N, --exclude-schema=SCHEMA
	do NOT dump the named schema(s)


src/Command/Db/DbDumpCommand.php
```diff
@@ -21,6 +21,8 @@ protected function configure()
         $this->setName('db:dump')
             ->setDescription('Create a local dump of the remote database');
         $this->addOption('schema', null, InputOption::VALUE_REQUIRED, 'The schema to dump. Omit to use the default schema (usually "main").')
+            ->addOption('pg-namespace', null, InputOption::VALUE_REQUIRED| InputOption::VALUE_IS_ARRAY, 'Dump the named namespace/schema(s) only (Postgresql specific)')
```
pjcdawkins, line 24: Is this different than --schema above?

digital2real: Yes : with postgresql schema != database (not like mysql or mariadb)
"A database contains one or more named schemas, which in turn contain tables. Schemas also contain other kinds of named objects, including data types, functions, and operators. The same object name can be used in different schemas without conflict; for example, both schema1 and myschema can contain tables named mytable. Unlike databases, schemas are not rigidly separated: a user can access objects in any of the schemas in the database they are connected to, if they have privileges to do so."
https://www.postgresql.org/docs/current/ddl-schemas.html

PR #1158, opened 2022-09-19
Fix env:delete command to respect --no-delete-branch in interactive mode

pjcdawkins:


src/Command/Environment/EnvironmentDeleteCommand.php
```diff
@@ -197,51 +198,67 @@ function ($environment) {
...
 
+        if ($needNewline) {
+            $this->stdErr->writeln('');
+        }
+
         if (empty($toDeleteBranch) && empty($toDeactivate)) {
-            if ($error) {
-                $this->stdErr->writeln('');
-            }
             $this->stdErr->writeln('No environment(s) to delete.');
```
CollierCZ, line 262: A minor thing, but I'd suggest this for clarity while you're in there already ([reasoning](https://developers.google.com/style/plurals-parentheses?hl=en)):
```suggestion
            $this->stdErr->writeln('No environments to delete.');
```

pjcdawkins: Agreed, thanks - although I want to make another PR with some wording changes for delete vs. deactivate as well

PR #1168, opened 2022-10-14
KYC Integration

hayleemillar:
Code changes required to verify a user's phone number with the KYC service before continuing with project creation. If a user is not verified, they are to be redirected to console's phone verification page.


hayleemillar:
@pjcdawkins Looking for final review for KYC launch.

PR #1174, opened 2022-10-26
Fix #1171: Autocompletion not working on Windows

### LINKED ISSUE #1171: Autocompletion not working on Windows with a fresh install
On Windows 10, using Git Bash, after the regular setup, the autocompletion outputs the following message:

```
$ platform bash: C:UsersIEUser.platformshbinplatform: command not found
```

It seems that the `\` character is stripped in the path.
Having a look at the generated `autocompletion.sh`, I figured out that the path to the `platform` CLI is not quoted, and the `\` are not doubled.

I was able to fix the issue by replacing:

```bash
RESULT=$(C:\Users\IEUser\.platformsh\bin\platform _completion --shell-type bash >/dev/null)
```

by:

```bash
RESULT=$("C:\Users\IEUser\.platformsh\bin\platform" _completion --shell-type bash >/dev/null)
```

After this change, and after sourcing `~/.bashrc`, the autocompletion works.
It ouputs another error, but probably related to the Windows Bash implementation:

```
__ltrim_colon_completions: command not found
```

lolautruche:
Add quotes around the application executable before passing it to the autocomplete hook script. 
Quoting prevents issues with the path when containing special characters (WS and/or `\`).

### DIFF
--- a/src/Command/Self/SelfInstallCommand.php
+++ b/src/Command/Self/SelfInstallCommand.php
@@ -102,7 +102,7 @@ EOT
         try {
             $args = [
                 '--generate-hook' => true,
-                '--program' => $this->config()->get('application.executable'),
+                '--program' => '"' . $this->config()->get('application.executable') . '"',
             ];
             if ($shellType) {
                 $args['--shell-type'] = $shellType;


pjcdawkins:
The fix might be

- this, or
- #1175 (using `--multiple` and thus `$1 _completion` in the hook), and/or
- https://github.com/stecman/symfony-console-completion/pull/92 (escaping the path and/or quoting the `$1`)

lolautruche:
Not sure what the `--multiple` option does. I read the description in the original command, but it's still unclear to me.

pjcdawkins:
The --multiple appears to just change `<path> _completion` to `$1 _completion` so if you run `platfoo <tab>` Bash should be using `platfoo` as the command name, etc. Which seems convenient, if it works.

lolautruche:
Thanks for the clarification 👍 
Seems handy indeed.

PR #1178, opened 2022-10-27
Update ssh-key commands to recommend certificates and avoid RSA

pjcdawkins:


src/Command/SshKey/SshKeyAddCommand.php
```diff
@@ -40,9 +45,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $email
         ));
 
+        $this->stdErr->writeln('<fg=yellow;options=bold>Notice</>');
+        $this->stdErr->writeln('SSH keys are no longer needed by default, as SSH certificates are supported.');
+        $this->stdErr->writeln('Certificates offer more security than keys.');
```
pjcdawkins, line 50:
```suggestion
        $this->stdErr->writeln('Certificates offer more features and better security than keys.');
```

PR #1181, opened 2022-10-31
📝 Add clarity to yes flag

CollierCZ:
It was unclear to a user that the `-y` flag would accept default options even if they weren't boolean. Hopefully, the added help text would help in such a situation.

Issue #1054, opened 2021-05-08
The -y/-n options are misleading


fabpot:
The global `-y` option is described like this: `Answer "yes" to any yes/no questions`, which is misleading IMHO (same for the `-n` one).

What it does is that it **only** answers **confirmation questions** (like confirming that you do want to `push` to master), but it does nothing on **option questions** that can also have yes/no answers (like `--sensitive` on `variable:create` for instance ).

But from a user perspective, the difference is very subtle:

```
Are you sure you want to push to the master (production) branch? [Y/n]
```

vs

```
Is the value JSON-formatted? [y|N]
```

For instance, when creating a variable, I don't want to answer all questions, the defaults are fine most of the time, but the only way to disable interaction is to use either `--yes` or `--no` as the Symfony `--no-interaction` option is not available for some reason. You can use either option as we are only interested in their side-effect here. I know that I can also set the `PLATFORMSH_CLI_NO_INTERACTION` environment variable, but that makes typing a lot with many possible typos (not even talking about Windows support :)).

I'm not sure how to improve over the current situation, but as a new user, I find the current behavior rather unintuitive.

pjcdawkins:
Agreed, it's a hangover from when it was mainly for Drupal users, who had Drush's equivalent -y/-n to contend with, and there was a clash between Symfony Console's -n (meaning "go ahead without interaction")  and Drush's -n (meaning "answer no to confirm questions", or "don't go ahead", more like a dry run)

pjcdawkins:
We could add `--no-interaction` in addition to the current options

And maybe look at modifying help output to show legacy vs current options

pjcdawkins:
Related PRs:

- https://github.com/platformsh/console-form/pull/9
- https://github.com/platformsh/platformsh-cli/pull/1182

pjcdawkins:
I believe this is fixed with https://github.com/platformsh/platformsh-cli/commit/42472e8b2b3c8c57f86243611e9cb077405e0327

PR #1182, opened 2022-10-31
Hide deprecated or confusing options from command help

pjcdawkins:
Follow up to #1181

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -2,6 +2,7 @@
 namespace Platformsh\Cli;
 
 use Platformsh\Cli\Console\EventSubscriber;
+use Platformsh\Cli\Console\HiddenInputOption;
 use Platformsh\Cli\Service\Config;
 use Platformsh\Cli\Util\TimezoneUtil;
 use Symfony\Component\Console\Application as ParentApplication;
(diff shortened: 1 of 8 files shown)


src/Application.php
```diff
@@ -66,8 +67,11 @@ protected function getDefaultInputDefinition()
             new InputOption('--quiet', '-q', InputOption::VALUE_NONE, 'Do not output any message'),
             new InputOption('--verbose', '-v|vv|vvv', InputOption::VALUE_NONE, 'Increase the verbosity of messages'),
             new InputOption('--version', '-V', InputOption::VALUE_NONE, 'Display this application version'),
-            new InputOption('--yes', '-y', InputOption::VALUE_NONE, 'Answer "yes" to confirmation questions; accept the default value for other questions; disable interaction'),
-            new InputOption('--no', '-n', InputOption::VALUE_NONE, 'Answer "no" to confirmation questions; accept the default value for other questions; disable interaction'),
+            new InputOption('--no-interaction', null, InputOption::VALUE_NONE, 'Do not ask any interactive questions'),
```
CollierCZ, line 70: Would it make sense to make it explicit that it accepts the defaults?
```suggestion
            new InputOption('--no-interaction', null, InputOption::VALUE_NONE, 'Do not ask any interactive questions; accept default values'),
```

CollierCZ: Also, I guess there's not a shorter version of the flag? Seems unfortunate to be able to shorten `yes` but not `no-interaction` :grinning:

pjcdawkins: That's the tricky part, `-n` is already a shortcut for `--no-interaction` in Symfony Console by default, but in this CLI it's the shortcut for `--no`.

In this PR the behavior would be:

- `--no-interaction`: accept default values (including for confirmation questions, whose default is usually yes but sometimes no)
- `--yes` / `-y`: answer yes to confirmation questions, otherwise accept default values
- `--no` / `-n`: answer no to confirmation questions, otherwise accept default values

pjcdawkins: With https://github.com/platformsh/platformsh-cli/pull/1183, I believe `-y` would have the same effect as `--no-interaction`.

pjcdawkins:
@CollierCZ after some thought I've updated this to:

1. hide --quiet/-q - it's confusing to me what "do not show any messages" should do, I'd expect it to hide only stderr messages and leave normal output, but it hides everything, and it's more UNIXy to redirect to /dev/null, and it was yet another option that was disabling interaction
2. add --no-interaction (for consistency with other Symfony apps as suggested in #1054)
3. keep --yes/-y there, and shown, as it is a useful shortcut for --no-interaction
4. hide --no but keep it working as described
5. show a warning if the `-n` shortcut for --no is used, as it doesn't have a purpose in scripting and it can be confused with --no-interaction (it's the default shortcut for --no-interaction in other Symfony apps)
6. add --ansi and --no-ansi to be more like other Symfony apps (they do work as described), but keep them hidden
7. hide deprecated options from various commands (they are only there for backwards compatibility)

PR #1189, opened 2022-11-21
Support slash style branches.

shahinam:
Currenly if there a slash in branch name like `feature/feature-name` the cli can not find the relevant git branch. 

Since git replaces the `/`  in branch name with `-` - we can easily do the same replacement and make it work for branches with slashes.

### DIFF
--- a/src/Service/Git.php
+++ b/src/Service/Git.php
@@ -101,7 +101,9 @@ class Git
     {
         $args = ['symbolic-ref', '--short', 'HEAD'];
 
-        return $this->execute($args, $dir, $mustRun);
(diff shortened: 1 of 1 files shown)


pjcdawkins:
> Currently if there a slash in branch name like feature/feature-name the cli can not find the relevant git branch.

Why / what makes you say that? I think slashes in branch names have worked for a long time?

shahinam:
Ahh, it seems it is because of CI - where in gitlab variable `CI_COMMIT_REF_SLUG` anything other than [0-9a-z] is transformed into a "-"

PR #1206, opened 2022-12-28
Allow the disk command to be used during crons/deployments

pjcdawkins:


src/Command/Metrics/DiskUsageCommand.php
```diff
@@ -128,14 +128,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $environment = $this->getSelectedEnvironment();
-        if (!$environment->isActive()) {
-            $this->stdErr->writeln(\sprintf('The environment %s is not currently active.', $this->api()->getEnvironmentLabel($environment, 'error')));
-            return 1;
-        }
 
         $environmentData = $environment->getData();
         if (!isset($environmentData['_links']['#metrics'])) {
-            $this->stdErr->writeln(\sprintf('The metrics API is not available on the environment: %s', $this->api()->getEnvironmentLabel($environment, 'error')));
+            $this->stdErr->writeln(\sprintf('The metrics API is not currently available on the environment: %s', $this->api()->getEnvironmentLabel($environment, 'error')));
```
matthiaz, line 134: I'm not a native english speaker, but shouldn't this be 'is currently not available' (not after currently)

pjcdawkins: Either seem OK to me (plus "is currently unavailable"), each with subtly different tone.

PR #1208, opened 2022-12-30
Add a command to list activity types locally

pjcdawkins:
Avoids needing to visit a page on the web


pjcdawkins:
Could be `activity:list --types` instead of a whole new command

PR #1209, opened 2023-01-04
Help users migrate to the new CLI

akalipetis:
* Add notification if both CLIs are installed
      If the user has both the New CLI and the Legacy CLI installed and the Legacy CLI has execution precendence because of the `PATH` config, a warning is displayed.
* Add migrate prompt for the new CLI
  The migrate prompt is disabled in the following cases:
  * CLIs for white-labels other than Platform.sh
  * CI systems
  * When the CLI is run in the new wrapper
* Better use config for custom Wrapped functionality

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -18,6 +18,7 @@ use Platformsh\Cli\Model\Host\RemoteHost;
 use Platformsh\Cli\Model\RemoteContainer;
 use Platformsh\Cli\Service\Shell;
 use Platformsh\Cli\Service\Ssh;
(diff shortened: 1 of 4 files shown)


src/Command/CommandBase.php
```diff
@@ -292,6 +297,11 @@ protected function interact(InputInterface $input, OutputInterface $output)
 
         $this->checkUpdates();
         $this->checkSelfInstall();
+        // Run migration steps only for the Platform.sh CLI
```
pjcdawkins, line 300:
```suggestion
        // Run migration steps if configured.
```

lolautruche (approved):

PR #1211, opened 2023-01-04
Installer: prompt to migrate to the wrapped CLI

pjcdawkins:


dist/installer.php
```diff
@@ -108,6 +118,21 @@ public function run() {
 
         $this->output($this->cliName . " installer", 'heading');
 
+        if ($this->migratePrompt && !getenv($this->envPrefix . 'WRAPPED') && !$this->isCI()) {
```
akalipetis, line 121: Is there any chance this runs with `PLATFORM_WRAPPED=1`?

pjcdawkins: Good point, no

akalipetis (approved):
LGTM

akalipetis (approved):
Even better with the option to cancel the wait by pressing enter. I think we're good!

PR #1217, opened 2023-01-09
Follow symlinks when checking for multiple CLI instances

### LINKED ISSUE #1215: The platform CLI seems to remove itself after upgrading from 3.88.1
Hello there, this morning I tried to initialize a new project using the CLI. Things didn't go quite as planned...
```
$ platform create
Checking for Platform.sh CLI updates (current version: 3.88.1)

Version 4.0.1 is available. Release notes:
[...]
Update to version 4.0.1? [Y/n] y
Updating to version 4.0.1
The Platform.sh CLI has been successfully updated to version 4.0.1
$ platform create

Warning: Both the Legacy (PHP) CLI and the New CLI are installed on your system.
To migrate fully, delete the Legacy CLI.

Remove the following file completely: /home/thomas/.platformsh/bin/platform
This operation is safe and doesn't delete any data.

Do you want to remove this file now? [Y/n] y
File successfully removed! Open a new terminal for the changes to take effect.
$ platform create
zsh: command not found: platform
```

I tried opening a new terminal, but that did not help. The `platform` binary is still missing 😢 

It seems confused by a symbolic link I have in `/usr/local/bin` (which is not technically needed since `~/.platformsh/bin` was added in my PATH, but anyway)
```
ls -la /usr/local/bin/platform
$lrwxrwxrwx 1 root root 37 Sep  8 18:07 /usr/local/bin/platform -> /home/thomas/.platformsh/bin/platform
```

Removing the symbolic link seems to fix the issue as the CLI does not ask to remove the file when I remove it, but I'd rather not change my PATH for every tool I setup on my machine, which is why I prefer to put things in `/usr/local`. Could the code detecting duplicate installations of the platform CLI ignore symbolic links (and maybe non-executable files as well) ?

Thanks in advance!

### LINKED ISSUE #1216: Migrate check should follow symlinks
If there's a symlink to the Platform.sh CLI, the new migrate check will get confused and prompt the user to delete the CLI.

Example reproduction output:

<details>

```console
➜  ~ platform auth:info
...
➜  ~ ln -s `which platform` ~/.local/bin/platform
➜  ~ platform auth:info

Warning: Both the Legacy (PHP) CLI and the New CLI are installed on your system.
To migrate fully, delete the Legacy CLI.

Remove the following file completely: /Users/akalipetis/.platformsh/bin/platform
This operation is safe and doesn't delete any data.

Do you want to remove this file now? [Y/n]
```

</details>

We should follow symlinks when checking for multiple binaries in the system.

pjcdawkins:
Closes #1215 and closes #1216 (duplicate), but this is a dirty fix

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -471,6 +471,7 @@ abstract class CommandBase extends Command implements MultiAwareInterface
         $appName = $config->get('application.executable');
         $paths = (new OsUtil())->findExecutables($appName);
         $pharPath = \Phar::running(false);
+        $paths = array_unique(array_filter(array_map('realpath', $paths)));
         $paths = array_diff($paths, [$pharPath]);
         if (count($paths)) {
             // Avoid deleting random directories in path
@@ -479,8 +480,8 @@ abstract class CommandBase extends Command implements MultiAwareInterface
                 return;
             }
 
-            // Check if the file is writable.
-            if (!is_writable($pharPath)) {
+            // Check if the path is a regular, writable file.
+            if (!is_file($pharPath) || !is_writable($pharPath)) {
                 return;
             }
 


akalipetis (approved):
I think this should work, I would follow the symlink as well for the .phar file, but I think there are a few (if any at all) cases where this would be an improvement.

LGTM

akalipetis:
@pjcdawkins I closed #1216 because it was a duplicate of #1215 (the underlying issue is the same) so please target that one instead.

PR #1222, opened 2023-01-27
Provide more ways to suppress the legacy warning

pjcdawkins:
Reads the following environment variables to suppress the warning about migrating to the new CLI installation method:
- PLATFORMSH_CLI_NO_LEGACY_WARNING=1
- IS_DDEV_PROJECT=true
- LANDO_INFO=anything

(in addition to the other currently supported variables which are `CI`, `BUILD_NUMBER` and `RUN_ID`)

Also accepts a command-line option `--no-legacy-warning`, e.g.

```sh
curl url/to/installer.php | php -- --no-legacy-warning
```


akalipetis (approved):
LGTM

PR #1224, opened 2023-02-02
Fix help command to properly display executable

akalipetis:
The help command did not extend CommandBase and thus the executable name was not interpolated correctly. Copied the help command over from the base help command to be the same, but extended from CommandBase for correct interpolation


akalipetis:
@pjcdawkins I can't put reviewers here, but feel free to take a look when you have time.

pjcdawkins:
@akalipetis You'll probably want to do the same for `list` to fix `platform help list`

akalipetis:
@pjcdawkins good catch, I don't there's any other left. Feel free to take another look.

Another change in the docs I made was related to the example namespace, instead of `platform list test` we now display `platform list project` which makes more sense IMO.

PR #1266, opened 2023-05-25
Add `deploy` alias to `environment:push` command

lolautruche:


pjcdawkins:
I think we can do this if `deploy` automatically implies `--activate`

We are looking at other ways to improve how `--activate` works - e.g. making the push and activate (or branch) into one action, using git push options.

Issue #1273, opened 2023-06-19
Replace deprecated phone verification endpoint


moritzschuh:
The `/me/phone` endpoint has been replaced with `/me/verification`, introducing the new verification method "ticket".
Existing requests to `/me/phone` need to be updated to make use of the new endpoint.
In case of "ticket" is specified as required verification method, a message needs to be shown to the user, asking them to reach out to support.

The return of the deprecated endpoint had the following format:
`{"verify_phone":false}`, requiring phone verification prior to project creation in case `verify_phone` was `true`.

The new endpoint return has the following format:
{"state":true,"type":"ticket"} User needs to reach out to a member of staff in order to create a project
{"state":true,"type":"phone"} User needs to verify their phone number in order to create a project (same as /me/phone: verify_phone:true)
{"state":false,"type":null} User is allowed to create projects without the need for verification (same as /me/phone: verify_phone:false)

PR #1275, opened 2023-06-20
Update calls to KYC verification endpoint

hayleemillar:
Contained in this pull request are updates regarding verification and the KYC endpoints called.  
  
1. To test, first set a user as suspicious:
```
POST https://api.platform.sh/kyc/<UUID>/suspicious
BODY: {"suspicious":true,"notes":"Testing KYC","level":2}
```
2. Log into the suspicious account in your browser
3. Run `bin/platform logout`
4. Then `bin/platform login`
Make sure that you are logged into and authenticate the account that you marked suspicious previously
5. Run `bin/platform project:create`

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -205,7 +205,7 @@ api:
   # Whether Centralized User Management is available (requires "auth" and "organizations").
   centralized_permissions: true
 
-  # Whether the Phone Verification APIs (/me/phone and /users/ID/phonenumber) are enabled (requires "auth").
+  # Whether the Phone Verification APIs (/me/verification and /users/ID/phonenumber) are enabled (requires "auth").
   phone_verification: true
 
   # Whether the Metrics API is enabled.
--- a/config.yaml
+++ b/config.yaml
@@ -206,7 +206,7 @@ api:
   centralized_permissions: true
 
   # Whether the Phone Verification APIs (/me/verification and /users/ID/phonenumber) are enabled (requires "auth").
-  phone_verification: true
+  user_verification: true
 
   # Whether the Metrics API is enabled.
   metrics: true
--- a/src/Command/Auth/VerifyPhoneNumberCommand.php
+++ b/src/Command/Auth/VerifyPhoneNumberCommand.php
@@ -105,11 +105,15 @@ class VerifyPhoneNumberCommand extends CommandBase
         });
 
         $this->debug('Refreshing phone verification status');
-        $needsVerify = $httpClient->post( '/me/phone?force_refresh=1')->json()['verify_phone'];
+        $needsVerify = $httpClient->post( '/me/verification?force_refresh=1')->json();
         $this->stdErr->writeln('');
 
-        if ($needsVerify) {
-            $this->stdErr->writeln('Phone verification succeeded but the status check failed.');
+        if ($needsVerify['state']) {
+            if ($needsVerify['type'] == 'phone') {
+                $this->stdErr->writeln('Phone verification succeeded but the status check failed.');
+            } else if ($needsVerify['type'] == 'ticket') {
+                $this->stdErr->writeln('Phone verification succeeded but staff verification is required. Please reach out to Support.');
+            }
             return 1;
         }
 
--- a/src/Command/Auth/VerifyPhoneNumberCommand.php
+++ b/src/Command/Auth/VerifyPhoneNumberCommand.php
@@ -112,7 +112,10 @@ class VerifyPhoneNumberCommand extends CommandBase
             if ($needsVerify['type'] == 'phone') {
                 $this->stdErr->writeln('Phone verification succeeded but the status check failed.');
             } else if ($needsVerify['type'] == 'ticket') {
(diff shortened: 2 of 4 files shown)


src/Command/Auth/VerifyPhoneNumberCommand.php
```diff
@@ -105,11 +105,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-        $needsVerify = $httpClient->post( '/me/phone?force_refresh=1')->json()['verify_phone'];
+        $needsVerify = $httpClient->post( '/me/verification?force_refresh=1')->json();
         $this->stdErr->writeln('');
 
-        if ($needsVerify) {
-            $this->stdErr->writeln('Phone verification succeeded but the status check failed.');
+        if ($needsVerify['state']) {
+            if ($needsVerify['type'] == 'phone') {
+                $this->stdErr->writeln('Phone verification succeeded but the status check failed.');
+            } else if ($needsVerify['type'] == 'ticket') {
```
pjcdawkins, line 114: hm, might phone verification have succeeded in this case? if so this command should return `0` (success).. unless we're changing the purpose of the command

moritzschuh: Yes, this command is specifically about phone verification, so it should ignore the ticket status.

pjcdawkins:
Fixes #1273

src/Command/Auth/VerifyPhoneNumberCommand.php
```diff
@@ -105,16 +105,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $this->stdErr->writeln('');
 
-        if ($needsVerify) {
-            $this->stdErr->writeln('Phone verification succeeded but the status check failed.');
-            return 1;
+        if ($needsVerify['state']) {
+            if ($needsVerify['type'] == 'phone') {
+                $this->stdErr->writeln('Phone verification succeeded but the status check failed.');
+                return 1;
+            }
```
pjcdawkins, line 115: Maybe this whole check is unnecessary, as this should never happen, right? and perhaps it will happen because computers... but the customer can't do much about it?

At least if we keep it I'd tweak it as follows, as I think the 'state' check wouldn't matter:

```suggestion
        if ($needsVerify['type'] == 'phone') {
            $this->stdErr->writeln('Phone verification succeeded but the status check failed.');
            return 1;
        }
```

Removing it would be deleting lines 106-116.

src/Service/Api.php
```diff
@@ -1425,19 +1425,14 @@ public function loadSubscription($id, Project $project = null, $forWrite = true)
      *
      * @return bool
```
pjcdawkins, line 1426:
```suggestion
     * @return array{'state': bool, 'type': string}
```

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -65,22 +65,31 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         // Check if the user needs phone verification before creating a project.
-        if ($this->api()->needsPhoneVerification()) {
-            $this->stdErr->writeln('Phone number verification is required before creating a project.');
-            if ($input->isInteractive()) {
-                $this->stdErr->writeln('');
-                $exitCode = $this->runOtherCommand('auth:verify-phone-number');
-                if ($exitCode !== 0) {
+        $needsVerify = $this->api()->needsVerification();
```
pjcdawkins, line 68: See below about renaming the method - the variable would need to be renamed accordingly

```suggestion
        $verificationStatus = $this->api()->checkVerificationStatus();
```

config.yaml
```diff
@@ -205,7 +205,7 @@ api:
   # Whether Centralized User Management is available (requires "auth" and "organizations").
   centralized_permissions: true
 
-  # Whether the Phone Verification APIs (/me/phone and /users/ID/phonenumber) are enabled (requires "auth").
+  # Whether the Phone Verification APIs (/me/verification and /users/ID/phonenumber) are enabled (requires "auth").
   phone_verification: true
```
pjcdawkins, line 209: Shall we rename this to `user_verification`?

src/Command/Project/ProjectCreateCommand.php
```diff
@@ -65,22 +65,31 @@ protected function configure()
...
                 }
-                $this->stdErr->writeln('');
-            } elseif ($this->config()->has('service.console_url')) {
-                $this->stdErr->writeln('');
-                $url = $this->config()->get('service.console_url') . '/-/phone-verify';
-                $this->stdErr->writeln('Please open the following URL in a browser to verify your phone number:');
-                $this->stdErr->writeln(sprintf('<info>%s</info>', $url));
-                return 1;
-            } else {
+            } else if ($needsVerify['type'] == 'ticket') {
```
pjcdawkins, line 86: We need an `else` block (in case the "type" is something new or unrecognized), and since an unrecognized verification type would also depend on Support at this stage, maybe this is simplest:

```suggestion
            } else { 
```

PR #1282, opened 2023-07-11
Remove confusing type info when branching

pjcdawkins:
Instead of:

> Creating a new environment `dev`, branched from `main` (type: production)

the `branch` command will now show:

> Creating a new environment `dev`, branched from `main`


pjcdawkins:
Superseded by #1283

PR #1297, opened 2023-07-31
Allow the default config file to be replaced

pjcdawkins:
Using the environment variable `CLI_CONFIG_FILE` the main configuration file can be specified, overriding the default: [`{CLI_ROOT}/config.yaml`](https://github.com/platformsh/legacy-cli/blob/main/config.yaml). This file is loaded before any other config and before the "user" config file and other environment variables.

This is to help with https://github.com/platformsh/cli/issues/81

We wouldn't want to recommend this for customer use, as it would be much harder to manage updates if the whole config was replaced at runtime.

### DIFF
--- a/src/Service/Config.php
+++ b/src/Service/Config.php
@@ -25,7 +25,11 @@ class Config
     {
         $this->env = $env !== null ? $env : $this->getDefaultEnv();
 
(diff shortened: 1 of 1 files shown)


akalipetis (approved):
LGTM 🐻

akalipetis (approved):
LGTM 🐻

PR #1298, opened 2023-07-31
Remove the recommendation to build after cloning

pjcdawkins:
Also fixes `get` when `local:build` is disabled


akalipetis (approved):
LGTM 🐻 

The only "breaking" change, is that it won't display that you can build the project when the build option is not used, but the `local:build` command is enabled. Given that the command is deprecated though, I think that's fine.

pjcdawkins:
That's an intentional change, yes

PR #1321, opened 2023-09-18
Loosely validate profile sizes in resources:set command

pjcdawkins:
Currently profile sizes must match exactly, with an exception of those prefixed by a dot ".5" being assumed as "0.5", etc.

This PR replaces this with PHP's loose comparison, which compares numeric strings like this as [floats](https://www.php.net/manual/en/language.types.float.php), meaning "4.0" matches "4" (and vice versa), ".5" still matches "0.5", and  "4.0000000000000001" also matches "4" (and an empty string would match "0", not that there is a size zero).

It has a side effect of making the code look a bit (I believe) simpler.

image

### DIFF
--- a/src/Command/Resources/ResourcesSetCommand.php
+++ b/src/Command/Resources/ResourcesSetCommand.php
@@ -381,16 +381,16 @@ class ResourcesSetCommand extends ResourcesCommandBase
     protected function validateInstanceCount($value, $serviceName, $service, $limit)
     {
         if ($service instanceof Service) {
-            return sprintf('The instance count of the service <error>%s</error> cannot be changed.', $serviceName);
+            throw new InvalidArgumentException(sprintf('The instance count of the service <error>%s</error> cannot be changed.', $serviceName));
         }
         $count = (int) $value;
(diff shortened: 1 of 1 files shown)


src/Command/Resources/ResourcesSetCommand.php
```diff
@@ -395,7 +390,7 @@ protected function validateInstanceCount($value, $serviceName, $service, $limit)
         if ($limit !== null && $count > $limit) {
             return sprintf('The instance count <error>%d</error> exceeds the limit %d.', $count, $limit);
```
akalipetis, line 391: Shouldn't this raise an `InvalidArgumentException`? Same at the 2 `return`s above.

akalipetis (changes requested):
`validateInstanceCount` should raise an exception, not `return` and error.

src/Command/Resources/ResourcesSetCommand.php
pjcdawkins: Indeed thanks!

akalipetis (approved):
LGTM 🐻

PR #1326, opened 2023-09-21
Simplify plans configuration for the project:create (create) command

pjcdawkins:
Remove service.available_plans config, so there is no static list of plans. The plans list is fetched either from the organization setup options (the `/organizations/{id}/setup/options` API, if `api.organizations` is enabled), or from the `/plans` API.


akalipetis (approved):
LGTM 🐻

PR #1333, opened 2023-10-02
Improve project:create (create) command

pjcdawkins:
- Hide timeout options in help
- Wait and retry while fetching the project
- Recommend the "get" command afterwards
- Avoid fetching /regions twice


bojanz:
Looks great!

PR #1338, opened 2023-10-11
Use MariaDB-specific command names in "db:" commands

pjcdawkins:
MariaDB introduced the `mariadb` command as a symlink in [version 10.4.6](https://mariadb.com/kb/en/mariadb-1046-release-notes/) and then made it a separate binary in [10.5.2](https://mariadb.com/kb/en/mariadb-1052-release-notes/) with `mysql` being a symlink. Apparently and unfortunately the Debian mariadb-client package doesn't include the `mysql` symlink; the `mariadb-client-compat` package is needed on top.

### DIFF
--- a/src/Command/Db/DbDumpCommand.php
+++ b/src/Command/Db/DbDumpCommand.php
@@ -209,8 +209,9 @@ class DbDumpCommand extends CommandBase
                 break;
 
             default:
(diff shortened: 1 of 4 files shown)


pjcdawkins:
This isn't possible because the relevant package version is on the application container and not easy enough to inspect

pjcdawkins:
Reopening with a hack:

```sh
command_name="$(command -v mariadb-dump || echo -n mysqldump)"; "$command_name" ...
```

PR #1339, opened 2023-10-11
Draft: 126 fix msging proj creation

hayleemillar:


src/Command/Project/ProjectCreateCommand.php
```diff
@@ -98,6 +99,27 @@ protected function execute(InputInterface $input, OutputInterface $output)
             try {
                 $organization = $this->validateOrganizationInput($input, 'create-subscription');
```
pjcdawkins, line 100: Hello, we are negotiating in Slack about possible changes to the HAL logic here so maybe there won't be a NoOrganizationsException

PR #1349, opened 2023-11-24
Lock while generating certificates or refreshing tokens

pjcdawkins:
This guards against most of the race condition bugs that occur while running
multiple instances of the CLI in parallel. It doesn't fix all of them.

### DIFF
--- a/src/Service/Api.php
+++ b/src/Service/Api.php
@@ -64,6 +64,9 @@ class Api
     /** @var TokenConfig */
     private $tokenConfig;
 
+    /** @var FileLock */
+    private $fileLock;
+
     /**
      * The library's API client object.
      *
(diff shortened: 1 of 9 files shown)


src/Service/Api.php
```diff
@@ -292,6 +311,63 @@ private function getConnectorOptions() {
...
+     *
+     * @return mixed|null
+     */
+    public function lock($lockName, callable $onWait = null, callable $check = null, $intervalMs = 300, $timeLimit = 15)
+    {
+        $runOnWait = false;
+        $cacheKey = 'lock:' . $lockName;
+        $start = \time();
+        while (\time() - $start < $timeLimit) {
+            $cached = $this->cache->fetch($cacheKey);
```
pjcdawkins, line 338: Unfortunately, although it's convenient, this should probably not use the cache because that can be disabled in config.

Issue #1354, opened 2023-12-01
Lock files apparently aren't working on OS X


pjcdawkins:
The locking in #1349 doesn't appear to have any effect on an OS X machine.

Perhaps a filesystem buffering issue.

pjcdawkins:
Mostly addressed in #1356, but it fails on the first run, if multiple simultaneous processes are run before the (empty) lock files are created.

PR #1356, opened 2023-12-05
Use flock and fflush for locking

pjcdawkins:
Attempt to address #1354


pjcdawkins:
On MacOS this appears to work _more_ of the time, but not all of the time.

PR #1362, opened 2023-12-11
Use MariaDB-specific command names in "db:" commands

pjcdawkins:


src/Service/Relationships.php
```diff
@@ -218,12 +218,30 @@ private function normalizeRelationships(array $relationships)
...
+     * See: https://jira.mariadb.org/browse/MDEV-21303
+     *
+     * @param array $database The database definition from the relationships.
+     * @return bool
+     */
+    public function supportsMariaDBCommands(array $database)
+    {
+        if (isset($database['type']) && (strpos($database['type'], 'mariadb:') === 0 || strpos($database['type'], 'mysql:') === 0)) {
+            list(, $version) = explode(':', $database['type'], 2);
+            return version_compare($version, '10.5', '>=');
```
pjcdawkins, line 233: this is the server version rather than the client version

but perhaps it's a good enough heuristic

pjcdawkins:
see #1338

PR #1369, opened 2023-12-12
Skip fetching SSH keys if a cert is present

pjcdawkins:
Further reduces calls to the `/me` API (after #1365 )


pjcdawkins:
Superseded by #1370

PR #1374, opened 2023-12-17
Use the new user-access API

pjcdawkins:
- [x] Depends on https://github.com/platformsh/platformsh-client-php/pull/88

### DIFF
--- a/src/Command/User/UserAddCommand.php
+++ b/src/Command/User/UserAddCommand.php
@@ -118,42 +118,47 @@ class UserAddCommand extends UserCommandBase
         }
 
         // Process the [email] argument.
+        // This can be an email address or a user ID.
+        // When adding a new user, it must be a valid email address.
         $email = null;
+        $update = stripos($input->getFirstArgument(), ':u');
         if ($emailOrId = $input->getArgument('email')) {
             $selection = $this->loadProjectUser($project, $emailOrId);
-            $update = stripos($input->getFirstArgument(), ':u');
-            if ($update && !$selection) {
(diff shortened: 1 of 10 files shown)


src/Command/User/UserAddCommand.php
```diff
@@ -118,63 +118,94 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                $selection = $this->loadProjectUser($project, $userId);
+                if (!$selection) {
+                    throw new InvalidArgumentException('User not found: ' . $userId);
                 }
-                $email = $questionHelper->choose($choices, 'Enter a number to choose a user to update:');
             } else {
                 $question = new Question("Enter the user's email address: ");
                 $question->setValidator(function ($answer) {
-                    return $this->validateEmail($answer);
+                    if (empty($value)) {
```
eflies, line 142: `empty($answer)`

src/Command/User/UserAddCommand.php
```diff
@@ -118,63 +118,94 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
 
         // Process the [email] argument.
-        $email = $input->getArgument('email');
-        if (!$email) {
+        $email = null;
+        if ($emailOrId = $input->getArgument('email')) {
```
eflies, line 122: Shouldn't we set `$email` somewhere in this block? Right now we are sending empty value on invitations

pjcdawkins: Yes, thanks

bojanz:
I've tested the new commands manually, and confirmed their requests using `-vvv`. So far so good.

PR #1380, opened 2023-12-29
Add pagination to the org:users command

pjcdawkins:
Adds `--count` and `--sort` options to the org:users command (`--count 0` to fetch all pages).

No specific `--page` option is provided as the API provides opaque cursor-based pagination which is powerful but could look confusing in this context.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "guzzlehttp/guzzle": "^5.3",
         "guzzlehttp/ringphp": "^1.1",
         "platformsh/console-form": ">=0.0.37 <2.0",
-        "platformsh/client": ">=0.79.4 <2.0",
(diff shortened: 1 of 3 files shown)


bojanz:
Notes from manual testing.

1) Doing `platform organization:users -o my-internal-org --count 0` gives me:
``` 
[RuntimeException]          
Member user info not found  
```              

2) When I use   `--page` to go to the next page, I am not told that I am on another page, and I do not get any text directing me to the previous page. I only see the usual:
```
More users are available (displaying 20, total 143)

Show all users with: --count 0
View the next page with: --page aWyIwMUdQTkNZMVdNOUJSUUs3TThWU0RUUVc0SiJd
```

That feels a bit confusing.

PR #1402, opened 2024-02-06
Fix metrics for multi-instance grid hosting

pjcdawkins:
Test for Upsun with:

```sh
# Download the normal Upsun CLI config
curl -s https://raw.githubusercontent.com/platformsh/cli/main/internal/config/upsun-cli.yaml > /tmp/upsun-cli.yaml

# Download the build for this PR
curl -s https://pr-1402-x5xx7sy-mpaw4nkgdwaea.eu-2.platformsh.site/platform.phar > /tmp/platform-pr-1402.phar

# Create an alias
alias upsun='CLI_CONFIG_FILE=/tmp/upsun-cli.yaml php /tmp/platform-pr-1402.phar'

# Run metrics commands
upsun cpu
upsun mem
```

### DIFF
--- a/src/Command/Metrics/MetricsCommandBase.php
+++ b/src/Command/Metrics/MetricsCommandBase.php
@@ -36,12 +36,12 @@ abstract class MetricsCommandBase extends CommandBase
     private $fields = [
         // Grid.
         'local' => [
-            'cpu_used' => "SUM((`cpu.user` + `cpu.kernel`) / `interval`, 'service')",
-            'cpu_percent' => "SUM(100 * (`cpu.user` + `cpu.kernel`) / (`interval` * `cpu.cores`), 'service')",
-            'cpu_limit' => "AVG(`cpu.cores`, 'service')",
+            'cpu_used' => "AVG(SUM((`cpu.user` + `cpu.kernel`) / `interval`, 'service', 'instance'), 'service')",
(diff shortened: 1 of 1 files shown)


src/Command/Metrics/MetricsCommandBase.php
```diff
@@ -36,12 +36,12 @@ abstract class MetricsCommandBase extends CommandBase
     private $fields = [
         // Grid.
         'local' => [
-            'cpu_used' => "SUM((`cpu.user` + `cpu.kernel`) / `interval`, 'service')",
-            'cpu_percent' => "SUM(100 * (`cpu.user` + `cpu.kernel`) / (`interval` * `cpu.cores`), 'service')",
-            'cpu_limit' => "AVG(`cpu.cores`, 'service')",
+            'cpu_used' => "AVG(SUM((`cpu.user` + `cpu.kernel`) / `interval`, 'service', 'instance'), 'service')",
```
ricardokirkner, line 39: why do you do AVG here? would it not be enough to just project the instance as well as the service?

pjcdawkins: This is the same as Console's `cpu.average.value` query. For this PR I want to fix what we are displaying which is one value per service, not one per instance; later we could add some kind of option to split up the instances. This is consistent with the display for Dedicated envs.

ricardokirkner (approved):
LGTM

PR #1404, opened 2024-02-07
Fix "Failed to extract zip" on Windows (by updating credential helpers)

pjcdawkins:
The docker-credential-helpers are now distributed as binaries rather than zip or tgz archives.

### DIFF
--- a/src/CredentialHelper/Manager.php
+++ b/src/CredentialHelper/Manager.php
@@ -178,15 +178,13 @@ class Manager {
             throw new \RuntimeException('Failed to download credentials helper from URL: ' . $helper['url']);
         }
 
-        // Work in a temporary file.
         $fs = new Filesystem();
-        $tmpFile = $fs->tempnam(sys_get_temp_dir(), 'cli-helper-extracted');
+
+        // Work in a temporary file.
+        $tmpFile = $destination . '-tmp';
         try {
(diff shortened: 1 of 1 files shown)


pjcdawkins:
The OS X version has signature problems

```
$ spctl -a ./.credential-helper 
./credential-helper: invalid signature (code or signature have been modified)
```

https://github.com/docker/docker-credential-helpers/issues/246

so I should restore the 0.6.4 version here

src/CredentialHelper/Manager.php
```diff
@@ -315,11 +343,13 @@ private function getHelper() {
             if (!$this->shell->execute('ldconfig --print-cache | grep -q libsecret')) {
                 throw new \RuntimeException('Unable to find a credentials helper for this system (libsecret is not installed)');
             }
+        }
 
-            return $helpers['secretservice'];
+        if (substr($helpers[$os][$arch]['url'], -4) === 'zip' && !class_exists('\\ZipArchive') && !$this->shell->commandExists('unzip')) {
```
pjcdawkins, line 348:
```suggestion
        if (substr($helpers[$os][$arch]['url'], -4) === '.zip' && !class_exists('\\ZipArchive') && !$this->shell->commandExists('unzip')) {
```

PR #1405, opened 2024-02-07
Support IPV6 ip address validation

hanoii:
I ended up testing this properly using the cli and all.

--

I haven't did a full tests of the project, I played with the code on

https://codesandbox.io/p/devbox/blissful-hill-56v4n9?file=%2Findex.php%3A33%2C1

### DIFF
--- a/src/Command/Environment/EnvironmentHttpAccessCommand.php
+++ b/src/Command/Environment/EnvironmentHttpAccessCommand.php
@@ -124,9 +124,20 @@ class EnvironmentHttpAccessCommand extends CommandBase
         if ($address == 'any') {
             return;
         }
-        $extractIp = preg_match('#^([^/]+)(/([0-9]{1,2}))?$#', $address, $matches);
-        if (!$extractIp || !filter_var($matches[1], FILTER_VALIDATE_IP) || (isset($matches[3]) && $matches[3] > 32)) {
+        $extractIp = preg_match('#^([^/]+)(/([0-9]{1,3}))?$#', $address, $matches);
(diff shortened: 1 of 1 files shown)


pjcdawkins:
Thanks @hanoii 

Closes #1399

pjcdawkins:
Would we also need to fix this bit above?

```php
        // Normalize the address so that we can compare accurately with the
        // current value returned from the API.
        if ($address == 'any') {
            $address = '0.0.0.0/0';
        } elseif ($address && !strpos($address, '/')) {
            $address .= '/32';
        }
```

hanoii:
I reviewed that code, and from what I gathered from our chats and my experience with the commands I think not.

That code only validates what comes from the API, and because your API doesn't truly expect an ipv6 address if you only specify the ipv6 ip it will come back with `/32`. Thinking a bit more, maybe it would be best to validate it against `/128` and at least detect it. but I think this should be handled before sending it to the server than after, I'll check if I can suggest something else.

hanoii:
I was wrong, fixing that too.

hanoii:
ci fails, but I assume that's not related to this PR, ready for review.

hanoii:
> I just realised the CI failures are because this is targeting `master`, it should be `main` - @hanoii would you mind rebasing and changing the PR target?

Done!

hanoii:
Is there a rought timeline into when 4.14.2 will get into a new official 5.x cli release?

pjcdawkins:
It ended up in release 5.0.10 last week

https://github.com/platformsh/cli/releases

Thanks again @hanoii

PR #1427, opened 2024-03-19
Add mfa_enabled and sso_enabled columns on the org:users command

pjcdawkins:


lsmith77:
would love to get this into prod quickly so that I can use it for my current re-occuring SOC2 evidence collection. thank you!

PR #1431, opened 2024-03-27
Use the API to validate if a service can have a disk

pjcdawkins:


src/Command/Resources/ResourcesCommandBase.php
```diff
@@ -45,28 +45,16 @@ protected function allServices(EnvironmentDeployment $deployment)
...
         }
-        // All services have a disk except 3 types.
-        if ($service instanceof Service) {
-            list($prefix) = explode(':', $service->type, 2);
-            $diskless = ['chrome_headless', 'memcached', 'redis', 'varnish'];
-            return !in_array($prefix, $diskless, true);
-        }
-        // Apps have a disk if they have mounts.
-        return !empty($service->mounts);
+        return isset($service->getProperties()['resources']['disk']['minimum']);
```
tejapsh, line 57: 1. It's resources.minimum.disk
2. atm, this will not be available in sizing_api_enabled=false projects

tejapsh: 3.  `resources.minimum.disk === 0` also means the service can use a disk.

pjcdawkins: 1. Updated thanks!
2. OK, the command checks for that
3. OK, this will return true if it is set to anything other than `null`.

PR #1432, opened 2024-04-03
Add --no-code option to backup:restore and deprecate --restore-code

pjcdawkins:


src/Command/Backup/BackupRestoreCommand.php
```diff
@@ -19,7 +19,10 @@ protected function configure()
             ->addArgument('backup', InputArgument::OPTIONAL, 'The ID of the backup. Defaults to the most recent one')
             ->addOption('target', null, InputOption::VALUE_REQUIRED, "The environment to restore to. Defaults to the backup's current environment")
             ->addOption('branch-from', null, InputOption::VALUE_REQUIRED, 'If the --target does not yet exist, this specifies the parent of the new environment')
-            ->addOption('restore-code', null, InputOption::VALUE_NONE, 'Whether code should be restored as well as data');
+            ->addOption('restore-code', null, InputOption::VALUE_NONE,
+                "Restore code as well as data."
+                . "\nNote: code would be restored to the environment but not to the Git repository; use Git to revert code permanently.")
```
tejapsh, line 24: nit: can we refer to app builds/slugs instead of code? That might make it easier to comprehend?

src/Command/Backup/BackupRestoreCommand.php
```diff
@@ -19,7 +19,10 @@ protected function configure()
...
+            ->addOption('no-restore-code', null, InputOption::VALUE_NONE, 'Do not restore code, only data.');
```
tejapsh, line 25: How about something like `--use-latest-code` instead?

PR #1435, opened 2024-04-11
Fix --gateway-ports option in tunnel commands

pjcdawkins:
Fixes https://github.com/platformsh/cli/issues/178 - a bug introduced in #1413 (versions: v4.15.3 / 5.0.11)


akalipetis (approved):
LGTM

PR #1447, opened 2024-05-21
Add commands to get and set build resources

pjcdawkins:
New commands:

* `resources:build:get` (aliased to `build-resources:get`)
* `resources:build:set` (aliased to `build-resources:set`)

Test for Upsun with:

```
# Download Upsun config
curl -s https://raw.githubusercontent.com/platformsh/cli/main/internal/config/upsun-cli.yaml > /tmp/upsun-cli.yaml

# Download the build for this PR
curl -s https://pr-1447-ryc5tsy-mpaw4nkgdwaea.eu-2.platformsh.site/platform.phar > /tmp/platform-pr-1447.phar

# Create an alias
alias upsun='CLI_CONFIG_FILE=/tmp/upsun-cli.yaml php /tmp/platform-pr-1447.phar'

# Then run the above commands as "upsun ..."
```

Test requirements are the same as the legacy CLI installation requirements:

> an operating system supported by PHP (Linux, OS X, or Windows) and PHP (5.5.9 or higher) installed, with the following extensions: curl, json, pcre, and phar.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "guzzlehttp/guzzle": "^5.3",
         "guzzlehttp/ringphp": "^1.1",
         "platformsh/console-form": ">=0.0.37 <2.0",
-        "platformsh/client": ">=0.82.1 <2.0",
+        "platformsh/client": ">=0.83.0 <2.0",
         "symfony/console": "^3.0 >=3.2",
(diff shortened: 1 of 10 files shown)


remilejeune2:
It's working as expected.
Could we use the `max_cpu` and `max_memory` values available in the `capabilities` to validate the input in the CLI?

PR #1461, opened 2024-06-28
Update backup:restore command resources options

pjcdawkins:
Closes https://github.com/platformsh/cli/issues/190

```
Usage:
 upsun backup:restore [--target TARGET] [--branch-from BRANCH-FROM] [--no-code] [--no-resources] [--resources-init RESOURCES-INIT] [-p|--project PROJECT] [-e|--environment ENVIRONMENT] [-W|--no-wait] [--wait] [--] [<backup>]

Arguments:
  backup                               The ID of the backup. Defaults to the most recent one

Options:
      --target=TARGET                  The environment to restore to. Defaults to the backup's current environment
      --branch-from=BRANCH-FROM        If the --target does not yet exist, this specifies the parent of the new
                                       environment
      --no-code                        Do not restore code, only data.
      --no-resources                   Do not override the target's existing resource settings.
      --resources-init=RESOURCES-INIT  Set the resources to use for new services: backup, parent, default or minimum.
                                       If not set, "backup" will be used.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "guzzlehttp/guzzle": "^5.3",
         "guzzlehttp/ringphp": "^1.1",
         "platformsh/console-form": ">=0.0.37 <2.0",
-        "platformsh/client": ">=0.84.0 <2.0",
+        "platformsh/client": ">=0.85.0 <2.0",
         "symfony/console": "^3.0 >=3.2",
         "symfony/yaml": "^3.0 || ^2.6",
         "symfony/finder": "^3.0",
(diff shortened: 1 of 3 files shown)


akalipetis (approved):
LGTM

pjcdawkins:
I missed a fairly important part - now added.

> Shall we query the Git version before including this option? How does the API respond on version 26 if we include it in the restore request?

Normally we want every server to be upgraded before releasing the CLI feature; it's just simpler. 

Those eager to use the API feature in the meantime can play with:

```sh
upsun project:curl 'backups/some-backup-ID/restore' --json '{"restore_resources": false}'
```

Issue #676, opened 2018-01-25
Allow to use the command through non-localhost environment


shavounet:
I've tried to use the CLI with docker, but I couldn't use some commands like `platform login` due to the php server being binded to localhost : 
https://github.com/platformsh/platformsh-cli/blob/8f2669e4d17aaee9cce94b335562b496e1e87421/src/Command/Auth/BrowserLoginCommand.php#L65

Ideally, in this place (and maybe others), it should be configurable (through an ENV variable for example).

*Test case* : create a simple image (ubuntu with php, git, ssh, ...), install the CLI inside the image, and start a new container with the command `platform login` and correct port mapping. It should display normal output, but when going on http://localhost:5000/ the connection is marked as closed by remote.

*Workaround* : I use an API token, but I don't know if there is an other place where this might lead to issues

pjcdawkins:
Security-wise, it absolutely needs to be on the local loopback interface (`127.0.0.1` rather than `localhost`), so that it doesn't cause requests over the network, and so that it can't be served by an external host (e.g. by a malicious attacker somehow taking over `localhost` via DNS).

The `platform auth:password-login` command is available for non-browser usage. That's also configurable as the default "login method", mentioned here: https://github.com/platformsh/platformsh-cli#customization

pjcdawkins:
The presence of `auth:password-login` and this upcoming help message should resolve this issue:
https://github.com/platformsh/platformsh-cli/commit/8baebb450776fc9fed22867bed25d1bb9f0b0083

PR #1524, opened 2025-02-17
Document containerized CLI configuration

PrinsFrank:
When installing the platform cli in docker (compose), the authentication flow still tries to redirect the user to 127.0.0.1:5000, even when that is not possible. There was an issue for this already in #676, but it took some effort to figure out what environment variable name was expected. As apparently I'm not the first one to run the cli in a docker container, this PR makes sure that the information on how to this is more readily available.


szepeviktor:
What are `[1]` and `[2]`?

PR #1540, opened 2025-06-12
Add read-only "automatic_deployments" to env:info

vitolkachova:
Add information about deployments mode to `environment:info` command. Read-only property (feel free to say it makes zero sense).
Drafted because referencing my fork of php-client lib.

### DIFF
--- a/src/Command/Environment/EnvironmentInfoCommand.php
+++ b/src/Command/Environment/EnvironmentInfoCommand.php
@@ -92,6 +92,10 @@ class EnvironmentInfoCommand extends CommandBase
             $headings[] = new AdaptiveTableCell($key, ['wrap' => false]);
             $values[] = $this->formatter->format($value, $key);
         }
+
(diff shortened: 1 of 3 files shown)


src/Command/Environment/EnvironmentInfoCommand.php
```diff
@@ -92,6 +92,10 @@ protected function listProperties(Environment $environment)
             $headings[] = new AdaptiveTableCell($key, ['wrap' => false]);
             $values[] = $this->formatter->format($value, $key);
         }
+
+        $headings[] = 'deployment_type';
+        $values[] = $environment->getSettings()->enable_manual_deployments ? 'manual' : 'automatic';
```
pjcdawkins, line 97:
```suggestion
        $values[] = $environment->getSettings()->getProperty('enable_manual_deployments', false) ? 'manual' : 'automatic';
```

... because it seems the property doesn't exist everywhere - it's only in git 32?

PR #1547, opened 2025-07-25
Add a command to display and set environment deployment type

vitolkachova:
I might not be a fan of the hint on my own

### DIFF
--- a/go-tests/go.mod
+++ b/go-tests/go.mod
@@ -4,7 +4,7 @@ go 1.24
 
 require (
 	github.com/go-chi/chi/v5 v5.2.2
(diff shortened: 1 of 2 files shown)


go-tests/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.24
 
 require (
 	github.com/go-chi/chi/v5 v5.2.2
-	github.com/platformsh/cli v0.0.0-20250512110214-68e4962f0990
+	github.com/platformsh/cli v0.0.0-20250731182252-f0fdaf7d1210
```
pjcdawkins, line 7: Needs update

PR #1550, opened 2025-08-04
Add validation case for setting manual deployment type

vitolkachova:


pjcdawkins:
I think we discussed removing this policy - closing for now

PR #1552, opened 2025-08-07
Cross-command reference between autoscaling and resources commands

ricardokirkner:
This requires PR https://github.com/platformsh/legacy-cli/pull/1551 to be merged first.

### DIFF
--- a/src/Command/Autoscaling/AutoscalingSettingsSetCommand.php
+++ b/src/Command/Autoscaling/AutoscalingSettingsSetCommand.php
@@ -34,7 +34,11 @@ class AutoscalingSettingsSetCommand extends CommandBase
             ->addEnvironmentOption()
             ->addAppOption();
 
-        $helpLines = ['Configure apps/workers for automatically scaling on an environment.'];
+        $helpLines = [
+            'Configure apps/workers for automatically scaling on an environment.',
+            '',
+            sprintf('You can also configure resources statically, by running: <info>%s resources:set</info>.', $this->config()->get('application.executable'))
+        ];
         if ($this->config()->has('service.autoscaling_help_url')) {
             $helpLines[] = '';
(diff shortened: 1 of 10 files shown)


src/Command/Resources/ResourcesSetCommand.php
```diff
@@ -50,7 +50,9 @@ protected function configure()
             '',
             sprintf('Profile sizes are predefined CPU & memory values that can be viewed by running: <info>%s resources:sizes</info>', $this->config()->get('application.executable')),
             '',
-            'If the same service and resource is specified on the command line multiple times, only the final value will be used.'
+            'If the same service and resource is specified on the command line multiple times, only the final value will be used.',
+            '',
+            'You can also configure your apps to automatically scale using the <info>autoscaling:set</info> command.'
```
pjcdawkins, line 55: This is good, but elsewhere, recommended commands are usually prefixed with `:` and the "executable" name, like the mention of `resources:sizes` just above-

```suggestion
            sprintf('You can also configure autoscaling, by running: <info>%s autoscaling:set</info>.', $this->config()->get('application.executable')),
```

src/Command/Autoscaling/AutoscalingSettingsSetCommand.php
```diff
@@ -0,0 +1,727 @@
...
+            ->addOption('instances-max', null, InputOption::VALUE_OPTIONAL, 'Maximum number of instances that will be scaled up to')
+            ->addOption('dry-run', null, InputOption::VALUE_NONE, 'Show the changes that would be made, without changing anything')
+            ->addProjectOption()
+            ->addEnvironmentOption()
+            ->addAppOption();
+
+        $helpLines = [
+            'Configure apps/workers for automatically scaling on an environment.',
+            '',
+            'For learning how to configure resources statically for your apps and services, refer to the <info>resources:set</info> command.'
```
pjcdawkins, line 40: See my other comment on using the `application.executable` config pattern here

PR #1553, opened 2025-08-13
Changing instance count on services where autoscaling is enabled is not allowed

ricardokirkner:
…ot allowed


src/Command/Resources/ResourcesSetCommand.php
```diff
@@ -140,6 +140,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
             }
         }
 
+        $autoscalingSettings = $this->api()->getAutoscalingSettings($environment)->getData();
+        $autoscalingEnabled = [];
+        foreach ($autoscalingSettings['services'] as $service => $serviceSettings) {
+            if (!empty($serviceSettings['enabled'])) {
+                $autoscalingEnabled[$service] = $serviceSettings['enabled'];
+            }
```
pjcdawkins, line 148:
```suggestion
            $autoscalingEnabled[$service] = !empty($serviceSettings['enabled']);
```

src/Command/Resources/ResourcesSetCommand.php
```diff
@@ -231,7 +237,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
             // Set the instance count.
             if (!$service instanceof Service) { // a Service instance count cannot be changed
-                if (isset($givenCounts[$name])) {
+                if (!empty($autoscalingEnabled[$name])) {
+                    $this->stdErr->writeln(sprintf('The instance count of the service <error>%s</error> cannot be changed when autoscaling is enabled.', $name));
```
pjcdawkins, line 241:
```suggestion
                    if (isset($givenCounts[$name])) {
                        $this->stdErr->writeln(sprintf('The instance count of the service <error>%s</error> cannot be changed when autoscaling is enabled.', $name));
                        $errored = true;
                    }
```

I'm not sure if we need a message here when nothing is specified - what do you think?

(And if something was specified, we should set `$errored`)

src/Command/Resources/ResourcesSetCommand.php
```diff
@@ -232,18 +238,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
                 } elseif ($showCompleteForm) {
                     $ensureHeader();
-                    $default = $properties['instance_count'] ?: 1;
-                    $instanceCount = $questionHelper->askInput('Enter the number of instances', $default, [], function ($v) use ($name, $service, $instanceLimit) {
-                        return $this->validateInstanceCount($v, $name, $service, $instanceLimit);
-                    });
-                    if ($instanceCount !== $properties['instance_count']) {
-                        $updates[$group][$name]['instance_count'] = $instanceCount;
+                    if (!empty($autoscalingEnabled[$name])) {
+                        $this->stdErr->writeln(sprintf('The instance count of the service <error>%s</error> cannot be changed when autoscaling is enabled.', $name));
```
pjcdawkins, line 254: I think we don't need this error, we can skip showing the instance count question entirely if autoscaling is enabled. Proposing a refactor in: https://github.com/platformsh/legacy-cli/pull/1556

PR #1568, opened 2025-09-23
Branding-related changes

pjcdawkins:
As [Platform.sh is now Upsun](https://upsun.com/platform-sh-is-now-upsun/), this removes many mentions of Platform.sh from the codebase and the default configuration. These follow from similar changes in the main/wrapper CLI (https://github.com/platformsh/cli/pull/275/files) which are in production.

The SVG image is the same one already configured in the main CLI for `upsun` and now `platform` too; the post-login page looks like this:
image


akalipetis (approved):
LGTM

The Config changes make sense only in the context of this being used as a standalone CLI, right?

